PR #4, opened 2021-11-02
Add `BLS` facilities

### LINKED ISSUE #3: Cryptography integration
The library currently lacks integration with a BLS/BLS12-381 library for the consensus cryptographic operations.

I'd suggest using `blst` as the "provider" library: https://github.com/supranational/blst

and then implementing the functionality under the phase 0 spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#crypto, in particular a `hash` function and the suite of `BLS` functions:

* `hash` (let's use the hash lib in `ssz_rs` for now)
* Sign
* Verify
* Aggregate
* FastAggregateVerify
* AggregateVerify
* KeyValidate

ralexstokes:
Fixes #3.


ralexstokes:
This may be a separate PR but an excellent way to verify this functionality will be passing the `bls` tests in #5.

Not sure if you want to bother with the test runners or not, that may be something I or someone else can do.

.gitignore
```diff
@@ -8,3 +8,6 @@ Cargo.lock
 
 # These are backup files generated by rustfmt
 **/*.rs.bk
+
+
+.idea/
```
ralexstokes, line 13: i tend to not check-in files that are specific to just my personal set-up... what do you think about dropping this to merge into `main`?

examples/bls.rs
```diff
@@ -0,0 +1,8 @@
+use ethereum_consensus::crypto;
```
ralexstokes, line 1: can likely drop this import if we only use `SecretKey` here

examples/bls.rs
```diff
@@ -0,0 +1,8 @@
...
+use ethereum_consensus::crypto::SecretKey;
+
+fn main() {
+    let sk = SecretKey::random();
+    let pk = sk.public_key();
+    println!("{:?}", pk);
+}
```
ralexstokes, line 8: and a minor thing but i tend to include trailing newlines -- what do you think?

src/crypto.rs
```diff
@@ -1,7 +1,165 @@
+use thiserror::Error;
 use ssz_rs::prelude::*;
+use rand::prelude::*;
+
+use blst::{min_pk as blst_core, BLST_ERROR};
+use crate::crypto::Error::SizeMismatch;
```
ralexstokes, line 6: one clippy complaint is this import which is not used. 

you can just remove it here. and in this file, we already have `Error` in scope so don't need to do anything special to use a variant

src/crypto.rs
```diff
@@ -1,7 +1,165 @@
...
+use blst::min_pk::AggregateSignature;
```
ralexstokes, line 7: and the other one is that this line is used -- below you reach it via `blst_core` name so it is unused here.

if you fix both of these things the CI should pass

src/crypto.rs
```diff
@@ -1,7 +1,165 @@
...
+
+pub const BLS_PUBLIC_KEY_BYTES_LEN: usize = 48;
+pub const BLS_SECRET_KEY_BYTES_LEN: usize = 32;
+pub const BLS_SIGNATURE_BYTES_LEN: usize = 96;
+
+pub type BLSPubkey = Vector<u8, BLS_PUBLIC_KEY_BYTES_LEN>;
+pub type BLSSignature = Vector<u8, BLS_SIGNATURE_BYTES_LEN>;
```
ralexstokes, line 14: will want to remove these once we decide how to unify the below representations with the SSZ functionality

src/crypto.rs
```diff
@@ -1,7 +1,165 @@
...
+        err == BLST_ERROR::BLST_SUCCESS
+    }
+}
+
+pub fn aggregate(signatures: &[Signature])-> Result<Signature, Error> {
+    if signatures.is_empty() {
+        return Err(Error::ZeroSizedInput);
+    }
+    let v: Vec<&blst_core::Signature> = signatures.into_iter().map(|x| &x.0).collect();
+    let s = blst_core::AggregateSignature::aggregate(&v[..], true)
```
ralexstokes, line 91:
```suggestion
    let s = blst_core::AggregateSignature::aggregate(&v, true)
```

this may just be a style thing bc i think the suggestion and what is here will end up producing the same thing -- do you have a reason for `&v[..]` over just `&v`?

.gitignore
dnkolegov: agree

examples/bls.rs
dnkolegov: fixed

examples/bls.rs
dnkolegov: agree

src/crypto.rs
dnkolegov: fixed

src/crypto.rs
dnkolegov: fixed

src/crypto.rs
dnkolegov: fixed

src/crypto.rs
dnkolegov: ok

src/crypto.rs
```diff
@@ -1,7 +1,192 @@
...
+        let res = sig.0.verify(true, msg, BLS_DST, avg, &pk, true);
+        res == BLST_ERROR::BLST_SUCCESS
+    }
+
+    pub fn validate(&self) -> bool {
+        let valid = self.0.validate();
+        match valid {
+            Ok(()) => true,
+            Err(_) => false
+        }
```
ralexstokes, line 67:
```suggestion
        self.0.validate().is_ok()
```

dnkolegov: ok

dnkolegov:
> @dnkolegov Great stuff! I had a small question. I'm writing some code to do the tests from `consensus-spec-tests` for the BLS signature aggregration. The test inputs are read in as a vector of strings, but your aggregation function will only accept a vector of type `Signature`. What's the right way to do this conversion?

Hi @pkopparla. I will add a conversion method.

pkopparla:
@dnkolegov Thank you! That will be a great help.

dnkolegov:
@pkopparla I added `from_bytes` for Signature. Does it work for you?

pkopparla:
@dnkolegov Yes it does, thank you!
Could I ask another simple question? I'm trying to read in public keys and convert them to the type `PublicKey` for which there is already a `from_bytes` method. So I try something like this:
```  
let a 
 = String::from("0xa491d1b0ecd9bb917989f0e74f0dea0422eac4a873e5e2644f368dffb9a6e20fd6e10c1b77654d067c0618f6e5a7f79a");
let pubkey = PublicKey::from_bytes(a.as_bytes()).unwrap();
```
but I get `BLST_BAD_ENCODING` error which doesn't explain anything. Would you know what I'm doing wrong here? I'm new to Rust.

ralexstokes:
@pkopparla you should try with the leading `0x` removed!

pkopparla:
> @pkopparla you should try with the leading `0x` removed!

Thanks! I think I need to use `hex::decode` rather than `String::from`. Is that right?

ralexstokes:
yeah, `hex::decode` will parse the hex string and provide the bytes you are looking for; it may even be tolerant to a leading `"0x"` prefix but keep this in mind if something fails

ralexstokes:
btw if you can push code, even if work-in-progress to your PR #7 it will likely be easier to go back and forth on stuff like this

pkopparla:
Got it, I'll do that shortly.

Issue #2, opened 2021-10-30
Refactor presets to use (full) const generics


ralexstokes:
Refer this commit https://github.com/ralexstokes/ethereum_consensus/commit/9de3638e353d3e19254ea9e38d71e916f4f5d7fb where the decision was made to go for a duplicated version of the spec for each "preset". The codebase will suffer code duplication in the short run but avoid the compile-time cost and unwieldly UX of highly generic types.

Ideally, we could avoid this by leveraging generics in Rust, but the const generics feature is not fully developed yet. I think when the Rust team ships something that closes this issue https://github.com/rust-lang/rust/issues/60551, we should check if we can do something like the following:

```rust
pub trait Preset {
    pub const ROOTS_LIMIT: usize;
    // ... other constants
}

pub struct Mainnet;
impl Preset for Mainnet {
    pub const ROOTS_LIMIT: usize = MAINNET_ROOTS_LIMIT;
        // ... other constants
}

pub struct BeaconState<P: Preset> {
    pub block_roots: Vec<Root, P::ROOTS_LIMIT>,
    // ... other fields
}

// then e.g. something like:
pub fn apply_block<P: Preset>(state: BeaconState<P>, block: BeaconBlock<P>, config: Config) -> BeaconState<P> {
    // ... implementation elided
}
```

ralexstokes:
This issue was partially resolve with #6 where the overhead of passing all of the required compile-time constants will be a bit of a chore *inside* the library, they can be specialized before users of this library have to deal with them which seems to be an acceptable compromise for now.

I'll leave this issue open as a reminder to revisit the ergonomics here once something like the above issue linked to `rust-lang/rust` is resolved.

PR #6, opened 2021-11-02
Make generic types in lieu of duplicating code for spec types with compile-time bounds

ralexstokes:
This is my current preference for (partially) resolving #2.

The best solution beyond this would be some way to bundle a batch of constants into a single generic type but it seems that the Rust compiler needs to implement more expressive const generics before this is possible (see further discussion in #2).

Duplicating the code is not great and I was already finding myself reaching for a macro to even keep the definitions of each preset in sync.

I had to add support for structs containing generics to `ssz_rs` which I have locally and will push a release tomorrow.

But wanted to get this up in case there was any feedback so far. This is a key decision on the structure of this repo and once it is made we can parallelize the work across many more threads :)


examples/sketch.rs
```diff
@@ -1,6 +1,11 @@
 use ethereum_consensus::altair::light_client::Snapshot;
+use ethereum_consensus::phase0::mainnet::{apply_block, BeaconState, SignedBeaconBlock};
 
 fn main() {
+    let state = BeaconState::default();
+    let signed_block = SignedBeaconBlock::default();
+    let _post = apply_block(state, signed_block);
```
ralexstokes, line 7: example usage: all generic bounds have been specified by the time an item is exposed in its respective `preset`.

src/phase0/presets/mainnet.rs
```diff
@@ -0,0 +1,96 @@
+pub use crate::domains::{DomainType, SigningData};
```
ralexstokes, line 1: the general approach is to re-export items in a given preset and *this* is where we can set the appropriate constants for the compile-time data that changes across preset.

things that aren't great:
1) still have _some_ code duplication across presets... i started writing a macro to resolve this so that the preset just looks like:

```rust
define_preset! {
    MAX_COMMITTEE_PER_SLOT: u64 = 64;
    TARGET_COMMITTEE_SIZE: u64 = 128;
    ...
    MAX_VOLUNTARY_EXITS: usize = 16; 
}
```
and think this is advisable after we have sketched out more of the state transition and have a better idea of how it will all fit together (bc debugging thru the macro could be annoying)

2) the types are a bit heavy w/ 6-8 const generics that have to flow up to the top-level operators of this crate.... given the above re: Rust's current feature set, this may be the best we can do right now...

src/phase0/state_transition.rs
```diff
@@ -0,0 +1,48 @@
+use crate::phase0::beacon_block::SignedBeaconBlock;
+use crate::phase0::beacon_state::BeaconState;
+
+pub fn apply_block<
```
ralexstokes, line 4: this is probably the worst it will get, so if we can handle this, then i'd say full-speed ahead :)

Issue #5, opened 2021-11-02
Verify this repo against the consensus spec tests


ralexstokes:
We want to verify that this code passes the consensus spec tests here: https://github.com/ethereum/consensus-spec-tests

There are two high-level prerequisites blocking the running of each test type:

- [x] Add the `consensus-spec-tests` repo as a git submodule
- [x] Extend this codebase to read the tests format and then integrate them (I'd suggest as Rust integration tests where possible, a bit like here https://github.com/ralexstokes/ssz_rs/tree/main/ssz_rs/tests)
    - refer to this for general information on the format: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats

And then we can track each type of test to run and/or debug:

- [x] [bls](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/bls)
- [ ] [epoch_processing](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/epoch_processing)
- [ ] [finality](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/finality)
- [ ] [fork_choice](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/fork_choice)
- [ ] [forks](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/forks)
- [ ] [genesis](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/genesis)
- [ ] [merkle](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/merkle)
- [ ] [operations](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/operations)
- [ ] [random](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/random)
- [ ] [rewards](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/rewards)
- [ ] [sanity](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/sanity)
- [ ] [shuffling](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/shuffling)
- [ ] [ssz_static](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/ssz_static)
- [ ] [transition](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/transition)

Links go to the individual formats/details for each test type.

ralexstokes:
part of the `bls` tests were done in #17.

there are more from later forks

[...]

PR #7, opened 2021-11-03
Added consensus-spec-tests submodule

pkopparla:
Addresses issue #5 

Added the `consensus-spec-tests` submodule under `.gitmodules` with option for git large filesystem handling under `.gitattributes`.

TBD: Reading and integrating tests.


pkopparla:
Hi Alex,

I wrote a single test for aggregating public keys. It fails in an interesting way, by only mismatching the first character out of 96 characters. Do you have any thoughts?

tests/bls_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+        })
+        .collect();
+
+    // println!("sigarray {:?}", sigarray);
+    let agg: AggregatePublicKey =
+        AggregatePublicKey::aggregate(&[&sigarray[0], &sigarray[1], &sigarray[2]], true).unwrap();
+    //println!("calculated aggregate {:?}", agg);
+    let outkey = PublicKey::from_aggregate(&agg);
+    let outkeystr = hex::encode(PublicKey::serialize(&outkey));
+    println!("this agg \n {:?}", &outkeystr[..96]);
```
ralexstokes, line 48: should only be 48 bytes here...

ralexstokes:
> I wrote a single test for aggregating public keys. It fails in an interesting way, by only mismatching the first character out of 96 characters. Do you have any thoughts?

Yeah this sounds like something to do w/ the encoding -- the top 3 bits of encoded BLS12-381 field elements are used as flags... perhaps something is getting mixed here... i can pull this branch and poke around to take a look

ralexstokes:
i'd suggest using the integrations from #11 as that will likely handle the encoding/decoding correctly and *i think* it will get around this problem 

and there is nothing to make this super clear but the one test type you picked is from `altair` functionality so we need to go build out some functionality around aggregating public keys... i'll try to do soon and let you know when that happens

pkopparla:
I tried writing a test for phase0 that checks a message is signed correctly, but it's not right yet. I suspect the `0x` might be causing issues again, but I'm not sure how to use the implementations from #11  that you mentioned. Would you have a suggestion on how to do this correctly?

Cargo.toml
```diff
@@ -14,6 +14,6 @@ blst = "0.3.6"
 rand = "0.8.0"
 thiserror = "1.0.30"
 sha2 = "0.9.8"
-hex = "0.4.3"
```
ralexstokes, line 17: lets leave this in :)

pkopparla: whoops will do, thought it was only being used for my old `hex::decode` stuff.

ralexstokes:
i'd suggest doing something along these lines:

https://github.com/ralexstokes/ethereum_consensus/commit/f8ee0985b667aa9b97ff35a02d65c61fed8a54e4


although when i run that test, it fails w/ a different signature which suggests there is a bug in the crypto!

ralexstokes:
and the bug should be fixed in #13. you'll need to rebase `main` once it is merged

pkopparla:
Great! Then I'll attempt to extend this to all the signing tests.

pkopparla:
Added tests for `bls/aggregate/small`. This also fails on the last case `aggregate_na_signatures` with `aggregation failed: ZeroSizedInput`.

pkopparla:
I've added a loop to go through all of the tests in `phase0/bls/sign/small/`. Most cases pass but the `sign_case_zero_privkey` case is failing on the secret key validation. Shall I start adding similar test loops for the other categories under `phase0/bls`?

pkopparla:
Closed in favor of new `testingv2` branch.

PR #12, opened 2021-11-05
wip: phase 0 spec implementation

dnkolegov:


src/primitives.rs
```diff
@@ -18,6 +18,7 @@ pub type Domain = Vector<u8, 32>;
 pub const GENESIS_SLOT: Slot = 0;
 pub const GENESIS_EPOCH: Epoch = 0;
 pub const FAR_FUTURE_EPOCH: Epoch = Epoch::MAX;
+pub const BASE_REWARDS_PER_EPOCH: usize = 4;
```
ralexstokes, line 21: this is _actually_ a phase 0 constant, so it is better imo to put it there...

could go in `src/phase0/mod.rs` even

src/primitives.rs
```diff
@@ -4,8 +4,8 @@ pub use ssz_rs::prelude::Root;
 
 pub type Slot = u64;
 pub type Epoch = u64;
-pub type CommitteeIndex = u64;
-pub type ValidatorIndex = u64;
+pub type CommitteeIndex = usize;
+pub type ValidatorIndex = usize;
```
ralexstokes, line 8: i think this makes sense.

hopefully we won't need to do anything silly to serialize as a `u64` one day

src/phase0/state_transition.rs
```diff
@@ -46,3 +56,134 @@ pub fn apply_block<
...
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    signed_block: SignedBeaconBlock<
+        MAX_PROPOSER_SLASHINGS,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        MAX_ATTESTER_SLASHINGS,
+        MAX_ATTESTATIONS,
+        MAX_DEPOSITS,
+        MAX_VOLUNTARY_EXITS,
+    >,
```
ralexstokes, line 92: the type signatures are going to be annoying huh

ralexstokes: i'd say we try to deal with the long form as see how bad it really does get, but one alternative that may be a bit more workable is to abbreviate:

```rust
pub fn verify_block_signature<
    const A: usize,
    const B: usize,
    const C: usize,
    const D: usize,
    const E: usize,
    const F: usize,
    const G: usize,
    const H: usize,
    const I: usize,
    const J: usize,
    const K: usize,
    const L: usize,
    const M: usize,
>(
    state: BeaconState<
        ...
    >,
    signed_block: SignedBeaconBlock<
        ...
    >,
```


and, we could always consider a macro here...

```rust
with_bounds! {
    pub fn verify_block_signature(state: BeaconState, signed_block: SignedBeaconBlock) {
    }
}
```

src/phase0/state_transition.rs
```diff
@@ -46,3 +56,134 @@ pub fn apply_block<
...
+    let epoch = if epoch.is_none() {
+        get_current_epoch(state)
+    } else {
+        epoch.unwrap()
+    };
+    let fork_version = if epoch < state.fork.epoch {
+        state.fork.previous_version.clone()
+    } else {
+        state.fork.current_version.clone()
+    };
```
ralexstokes, line 144: can skip the clone here if `compute_domain` takes a reference
```suggestion
    let fork_version = if epoch < state.fork.epoch {
        &state.fork.previous_version
    } else {
        &state.fork.current_version
    };
```

src/phase0/state_transition.rs
```diff
@@ -46,3 +56,134 @@ pub fn apply_block<
...
+
+    return compute_domain(
+        domain_type,
+        fork_version,
+        state.genesis_validators_root.clone(),
```
ralexstokes, line 149: same, avoid clone

src/phase0/state_transition.rs
```diff
@@ -46,3 +56,134 @@ pub fn apply_block<
...
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+) -> Epoch {
+    todo!()
+}
+
+pub fn compute_signing_root<T: SimpleSerialize>(
+    ssz_object: T,
+    domain: Domain,
+) -> Result<ssz_rs::Root, ssz_rs::MerkleizationError> {
```
ralexstokes, line 180: let's use the `Root` from `crate::primitives`, they point to the same thing but pulling from the local module has a bit more semantic weight in the right direction

ralexstokes: and i'd also suggest extending your `Error` type to handle an error from the `MerklizationError` variety.

`thiserror` has a handy `from(...)` attribute in case you haven't seen it

src/phase0/state_transition.rs
```diff
@@ -46,3 +56,134 @@ pub fn apply_block<
...
+    let context = MerkleizationContext::new();
+    let object_root = ssz_object.hash_tree_root(&context)?;
+    let sd = SigningData {
+        object_root,
+        domain,
+    };
+    let context = MerkleizationContext::new();
```
ralexstokes, line 187: ah can re-use this here!

i'm not sure how to communicate this, but i'll go add docs...

for now, it is just read-only

ralexstokes: a possible future version of this code will actually just pass around a single `MerkleizationContext`

ralexstokes (commented):
great start! left some thoughts/comments below :)

src/primitives.rs
dnkolegov: fixed

src/phase0/state_transition.rs
ralexstokes: if you rebase onto `main` https://github.com/ralexstokes/ethereum_consensus/commit/206667028bc6da46b08e210577d2a2660c0aa14d

then you should see an `Error` type exposed in the `ssz_rs::prelude`. and I just realized there will be name collisions.

so you'll need something like:

```rust
use ssz_rs::prelude::*;
use ssz_rs::prelude::Error as SSZError;
```

and we should probably just expose the `ssz_rs` top-level error as that name, but this will work for now

src/phase0/state_transition.rs
dnkolegov: fixed

src/phase0/state_transition.rs
dnkolegov: fixed

src/phase0/state_transition.rs
dnkolegov: fixed

src/phase0/state_transition.rs
dnkolegov: fixed

src/domains.rs
```diff
@@ -4,13 +4,27 @@ use ssz_rs::prelude::*;
...
+            DomainType::BeaconAttester => 1u32.to_le_bytes(),
+            DomainType::Randao => 2u32.to_le_bytes(),
+            DomainType::Deposit => 3u32.to_le_bytes(),
+            DomainType::VoluntaryExit => 4u32.to_le_bytes(),
+            DomainType::SelectionProof => 5u32.to_le_bytes(),
+            DomainType::AggregateAndProof => 6u32.to_le_bytes(),
+        }
+    }
+}
+
```
ralexstokes, line 27: what do you think about something like this?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=744eaf45903e77bce6c7b41d2889d8e8

src/phase0/configs/mainnet.rs
```diff
@@ -1,8 +1,10 @@
 use crate::primitives::{Epoch, Gwei, VersionBytes};
+use ssz_rs::Vector;
 
 pub const MIN_GENESIS_ACTIVE_VALIDATOR_COUNT: u64 = 16384;
 pub const MIN_GENESIS_TIME: u64 = 1606824000;
 pub const GENESIS_FORK_VERSION_BYTES: VersionBytes = [0u8; 4];
+pub const GENESIS_FORK_VERSION: Vector<u8, 4> = Vector::from_iter(GENESIS_FORK_VERSION_BYTES);
```
ralexstokes: i had issues w/ this bc `from_iter` is not `const`... did I miss something? i'd prefer this if it works!

src/phase0/configs/mainnet.rs
```diff
@@ -1,14 +1,16 @@
 use crate::primitives::{Epoch, Gwei, VersionBytes};
+use ssz_rs::Vector;
 
 pub const MIN_GENESIS_ACTIVE_VALIDATOR_COUNT: u64 = 16384;
 pub const MIN_GENESIS_TIME: u64 = 1606824000;
 pub const GENESIS_FORK_VERSION_BYTES: VersionBytes = [0u8; 4];
```
ralexstokes, line 6: we can drop `GENESIS_FORK_VERSION_BYTES` if we figure out how to declare `GENESIS_FORK_VERSION: Version` as a constant

src/phase0/state_transition/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    domain_type: DomainType,
+    epoch: Option<Epoch>,
+) -> Result<Domain, Error> {
+    let epoch = if epoch.is_none() {
+        get_current_epoch(&state)
+    } else {
+        epoch.unwrap()
+    };
```
ralexstokes, line 283:
```suggestion
    let epoch = epoch.unwrap_or_else(|| get_current_epoch(&state));
```

src/phase0/state_transition/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+    let fork_version = if epoch < state.fork.epoch {
+        Some(state.fork.previous_version.clone())
+    } else {
+        Some(state.fork.current_version.clone())
+    };
```
ralexstokes, line 288: why did the `clone`s come back?

src/domains.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+pub fn compute_activation_exit_epoch(epoch: Epoch) -> Epoch {
+    epoch + 1 + MAX_SEED_LOOKAHEAD
+}
+
+pub fn compute_fork_digest(
+    current_version: Version,
+    genesis_validators_root: Root,
+) -> Result<ForkDigest, Error> {
+    let fork_data_root = compute_fork_data_root(current_version, genesis_validators_root)?;
+    let fork_digest: Vector<u8, 4> = ForkDigest::deserialize(fork_data_root.as_ref())?;
```
ralexstokes, line 353: using `deserialize` here should actually always error bc the input will be longer than the possible length of 4 bytes; and also we know the root is in fact a `Root` so no need to deserialize, which in general may be some strange pathological input that needs special care to internalize.

so instead, i think here we can do something like:
```rust
let fork_digest = fork_data_root[..4].try_into().expect("is the correct length");
```

where we want to ensure that a `ForkDigest` is just an alias for `[u8;4]`. the latest version of `ssz_rs` should have support for arrays of bytes like this.

src/phase0/configs/mainnet.rs
dnkolegov: what about just
`pub const GENESIS_FORK_VERSION: [u8; 4] = [0u8; 4];`

src/primitives.rs
```diff
@@ -31,6 +31,20 @@ impl fmt::Debug for Bytes32 {
...
+        let mut result = Vector::default();
+        for (i, (a, b)) in self.0.iter().zip(other.0.iter()).enumerate() {
+            result[i] = a ^ b;
+        }
+        Self(result)
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        self.0.as_slice()
+    }
```
ralexstokes, line 45: i haven't looked more deeply into the specifics but i'd *like* to avoid the need for methods like this, even if it means keeping in the inner type `pub` in the crate... i'm not completely sure we can around it though (e.g. if a user of this crate needs the underlying `&[u8]` some day for a `Bytes32`...

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: because now it's the fastest way to prototype all primitives and then improve their API's.
If remove `clone()` we have to define all API first.

src/phase0/configs/mainnet.rs
dnkolegov: I think you haven't pulled all updates

src/phase0/state_transition/mod.rs
dnkolegov: yes, it works with the latest ssz_rs. I was not aware of the changes.
Fixed.

src/primitives.rs
```diff
@@ -3,11 +3,11 @@ use std::fmt;
 
 pub use ssz_rs::prelude::Root;
 
-pub type Slot = u64;
-pub type Epoch = u64;
-pub type CommitteeIndex = u64;
-pub type ValidatorIndex = u64;
-pub type Gwei = u64;
+pub type Slot = usize;
+pub type Epoch = usize;
```
ralexstokes, line 7: something to consider here is that `Slot` and `Epoch` will _definitely_ get serialized and if this code runs on a machine where the word size is not 64bits then that machine will not be able to follow the consensus....

i'd lean towards keeping these as `u64` and if at some point they need to be converted into `usize`, e.g. they get pulled into some bounds computations, then we can cast there

src/phase0/configs/mainnet.rs
```diff
@@ -1,14 +1,16 @@
 use crate::primitives::{Epoch, Gwei, VersionBytes};
+use ssz_rs::Vector;
```
ralexstokes, line 2: does this need to be here?

src/phase0/configs/mainnet.rs
ralexstokes: yeah i think this was before i added the `[u8;N]` `SimpleSerialize` impls, agree that we can drop the `GENESIS_FORK_VERSION_BYTES` thing (that was just going to be a workaround)

src/phase0/configs/mainnet.rs
dnkolegov: fixed

src/primitives.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: It doesn't work

ralexstokes:
NOTE: CI failed due to "unused functions" warnings from clippy -- going to merge as there is more on top of this PR but want to keep this PR from getting any bigger

PR #17, opened 2021-11-09
BLS tests with error handling

pkopparla:
Addresses #5 

Tests for five phase0 BLS functions: `sign`,`verify`,`aggregate`,`aggregate_verify` and `fast_aggregate_verify`. 

At the moment, the `fast_aggregate_verify` fails on `consensus-spec-tests/tests/general/phase0/bls/fast_aggregate_verify/small/fast_aggregate_verify_infinity_pubkey` by verifying `true` when expected to be `false`. Possibly a bug?

Other functions passing all tests.


ralexstokes (commented):
general request to format code with `rustfmt` -- should be able to do `cargo fmt` from the project root directory to get the updated formatting

dnkolegov:
@pkopparla please look at https://github.com/ralexstokes/ethereum_consensus/pull/18
That PR should fix the issues you reported

pkopparla:
> @pkopparla please look at #18 That PR should fix the issues you reported

Thanks @dnkolegov , I'll try again when that is merged.

ralexstokes:
@pkopparla you can rebase this branch on main and it should include the fix mentioned above

pkopparla:
@ralexstokes I'm getting an error on compilation after merging the latest changes: 
```
 --> src/phase0/state_transition/context.rs:7:5
  |
5 | #[derive(Debug, Default)]
  |          ----- in this derive macro expansion
6 | pub struct Context {
7 |     merkleization_context: MerkleizationContext,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `ssz_rs::MerkleizationContext` cannot be formatted using `{:?}` because it doesn't implement `Debug`
```
Does it work for you?

ralexstokes:
@pkopparla yeah i had an issue where `cargo` didn't pick up the new git head for `ssz_rs`

i think you can do `cargo clean && cargo build` to see if that updates the local registry

pkopparla:
Thanks @ralexstokes `cargo update` did the trick. All the tests are now passing. I could move onto a new set of tests, but I had a few questions on making the code cleaner:
1. Does it make sense to split each of the 5 tests into its own `.rs` file? Or leave as is?
2. Is the error handling ok? Should it be more specific?
3. Should I be using any other rust features to reduce code repetition? macros? enums?
4. Any other comments on code/organization/documentation so far?

I understand it's a balance between "it works, let's keep moving" and "let's take a bit more time write readable code". But I'm still new to open source software and just wondering know how it's done.

ralexstokes:
excellent @pkopparla !

i'll review what is here and as we go back and forth on review, i think we will address most of your questions

Cargo.toml
```diff
@@ -6,6 +6,9 @@ license = "MIT OR Apache-2.0"
 
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+default = ["ef_spec_tests"]
+ef_spec_tests = []
```
ralexstokes, line 11: nitpick but i'd rather this just be `spec_tests` with the `ef` dropped :)

tests/README.md
```diff
@@ -0,0 +1,25 @@
+# Tests from Ethereum Consensus Spec Tests
```
ralexstokes, line 1: love more documentation :)

tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+The consensus spec tests repo is [here](https://github.com/ethereum/consensus-spec-tests).
+
+This repo has been added as a submodule. The test data is stored using the Github Large File System (LFS) and will not be automatically downloaded if you just clone the current repo. To get the test data you have to do
+
+```
+git lfs install
+git lfs pull
+``` 
+
+The tests are currently behind a feature `ef_spec_tests` which is enabled by default in `Cargo.toml`. If you don't want to run these tests, just turn it off there.
```
ralexstokes, line 12:
```suggestion
The tests are currently behind a feature `ef_spec_tests` which is enabled by default in `Cargo.toml`. If you don't want to run these tests, you can run `cargo` with the default features disabled.
```

tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+The description of the specs for this category can be found [here](https://github.com/ethereum/consensus-specs/tree/master/tests/formats/bls).
+
+Currently there are 5 types of test data for different functions: `sign`, `verify`, `aggregate`, `aggregate_verify` and `fast_aggregate_verify`. The basic testing pattern is as follows:
+
+1. Define a custom input-output struct that matches the test data organization. The output part is one variable but the input part uses another struct to capture different input fields (publickeys, messages, signatures).
+
+2. Read-in the test data using `serde_yaml` and parse it into the struct while accounting for hexadecimal encoding with `hex::decode`.
+
+3. Apply the function of interest (imported from [crypto.rs](https://github.com/ralexstokes/ethereum_consensus/blob/main/src/crypto.rs) and implemented for the input-output struct) and assert that the function output matches the listed test data output.
```
ralexstokes, line 25: this is the kind of thing that takes a lot of work (there are more test types beyond `bls`) to maintain/synchronize and it may be useful to help orient your work but i'd suggest dropping this and the implicit commitment to document each test type this way.

if you want to add some pointers to additional info, i'd just point readers to the `format`s part of the spec: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
+use ethereum_consensus::crypto::{
```
ralexstokes, line 1: i think one file is fine here imo

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+use std::fmt::Debug;
+use std::fs::File;
+
+fn decode_hex_with_prefix<T: AsRef<[u8]> + Debug>(data: T) -> Vec<u8> {
+    // a small check for empty strings
+    if data.as_ref().len() >= 2 {
+        hex::decode(&data.as_ref()[2..]).expect("is well-formed hex")
+    } else {
+        hex::decode(&data.as_ref()).expect("is well-formed hex")
+    }
```
ralexstokes, line 17: something that may be more tolerant to a wider variety of inputs is to just check if there is a prefix `0x` using e.g. `String::starts_with` https://doc.rust-lang.org/stable/std/string/struct.String.html#method.starts_with

something like:
```rust
let data = data.as_ref();
let data = if data.starts_with("0x") {
    &data[2..]
else {
    data
};
hex::decode(data).expect("...")
```


if the empty string is a special case worth checking specifically, you could check `data.is_empty()`

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+
+impl SigningTestIO {
+    fn verify_signature(&self) -> bool {
+        let input = &self.input;
+
+        let secret_key_bytes = decode_hex_with_prefix(&input.privkey);
+        let secret_key = match SecretKey::from_bytes(&secret_key_bytes) {
+            Ok(sk) => sk,
+            Err(_) => {
+                // this is the case empty secret key case
```
ralexstokes, line 40: which test case has this? i want to double-check the `~` isn't some yaml encoding

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+                if self.output == String::from("~") {
+                    return true;
+                } else {
+                    return false;
+                }
```
ralexstokes, line 45: just `return self.output == "~"`

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+
+        let signature_bytes = decode_hex_with_prefix(&self.output);
+        let expected_signature = Signature::from_bytes(&signature_bytes).unwrap();
+
+        signature == expected_signature
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+
```
ralexstokes, line 60: formatting: these empty line should disappear b/t the struct defn and the attributes that apply to it

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+            .collect();
+        let pubkeys = match pubkeys_result {
+            Ok(pk) => pk,
+            // error handling for infinity pub key
+            Err(_) => {
+                if self.output == false {
+                    return true;
+                } else {
+                    return false;
+                }
```
ralexstokes, line 121: ditto on simplifying this conditional

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+    output: bool,
+}
+
+impl AggVerifyTestIO {
+    fn aggregate_verify(&self) -> bool {
+        let pubkeys_result: Result<Vec<PublicKey>, _> = self
+            .input
+            .pubkeys
+            .iter()
+            .map(|x| PublicKey::from_bytes(&decode_hex_with_prefix(&x)))
```
ralexstokes, line 111: do you need this ref on `x`?
```suggestion
            .map(|x| PublicKey::from_bytes(&decode_hex_with_prefix(x)))
```

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+        let signature: Signature =
+            match Signature::from_bytes(&decode_hex_with_prefix(&self.input.signature)) {
+                Ok(sign) => sign,
+                Err(_) => {
+                    // handling for the zero signature case which raises a blst bad encoding error
+                    if self.output == false {
+                        return true;
+                    } else {
+                        return false;
+                    }
```
ralexstokes, line 140: ditto on handling here

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+                }
+            };
+        let verify_result = aggregate_verify(&pubkeys, messages_slice.as_ref(), signature);
```
ralexstokes, line 143: i'd say this is slightly more preferred in rust land
```suggestion
        let verify_result = aggregate_verify(&pubkeys, &messages_slice, signature);
```

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+        verify_result == self.output
+    }
+}
+#[derive(Debug, Serialize, Deserialize)]
+
```
ralexstokes, line 148: whitespace

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+                    }
+                }
+            };
+        let verify_result =
+            fast_aggregate_verify(pubkeys_ref.as_slice(), message_slice, &signature);
+        verify_result == self.output
+    }
+}
+#[derive(Debug, Serialize, Deserialize)]
+
```
ralexstokes, line 199: whitespace

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+struct VerifyInput {
+    pubkey: String,
+    message: String,
+    signature: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+
```
ralexstokes, line 207: whitespace

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+                    serde_yaml::from_reader(file).expect("is well-formatted yaml");
+                assert!(test_case.verify_aggregate())
+            }
+            Err(e) => println!("{:?}", e),
+        }
+    }
+}
+#[test]
+#[cfg(feature = "ef_spec_tests")]
+
```
ralexstokes, line 325: whitespace

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+                assert!(test_case.aggregate_verify())
+            }
+            Err(e) => println!("{:?}", e),
+        }
+    }
+}
+
+#[test]
+#[cfg(feature = "ef_spec_tests")]
+
```
ralexstokes, line 306: whitespace

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+                    } else {
+                        return false;
+                    }
+                }
+            };
+        let verify_result = signature.verify(&pubkey, message);
+        verify_result == self.output
+    }
+}
+
```
ralexstokes, line 245: i'd recommend putting all of the tests here into a `tests` mod, and then you can apply the feature flag once

(note: i may have the syntax a bit off for the combined `cfg` value)

```rust
#[cfg(test, feature = "ef_spec_tests")]
mod tests {
   #[test]
    fn test_...() { ... }
    // etc.
}
```

ralexstokes (commented):
this is an excellent start!

i went through and left some feedback...

there are a few other things i want to discuss but need to run for the moment so i'll leave this (partial) review for now

ralexstokes (commented):
one more thing i'd consider is moving the data parsing into `serde`

there are field attributes you can use to specify the desired transformation and that should keep the function bodies smaller and more focused by moving the parsing code elsewhere...

to give an example, `SigningTestIO` would look like:

```rust
#[derive(Debug, Serialize, Deserialize)]
struct SigningInput {
    #[serde(deserialize_with = "try_decode_hex_with_prefix")]
    privkey: Option<SecretKey>,
    #[serde(deserialize_with = "decode_hex_with_prefix")]
    message: Vec<u8>,
}

#[derive(Debug, Serialize, Deserialize)]
struct SigningTestIO {
    input: SigningInput,
    #[serde(deserialize_with = "decode_hex_with_prefix")]
    output: Vec<u8>,
}
```

where you need to add a "fallible" version of `decode_hex_with_prefix` that would return an `None: Option<SecretKey>` if the key fails to parse.

you would also need to adapt the function signature of `decode_hex_with_prefix` to match the expected signature, here is an example: https://stackoverflow.com/a/46755370 (under the first header with their fn `from_hex`).

and this would be adjusted for each test type...

tests/bls_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+        glob("consensus-spec-tests/tests/general/phase0/bls/aggregate_verify/small/**/*.yaml")
+            .expect("Failed to read glob pattern")
+    {
+        match entry {
+            Ok(path) => {
+                println!("{:?}", path.display());
+                let file = File::open(path).expect("file exists");
+                let test_case: AggVerifyTestIO =
+                    serde_yaml::from_reader(file).expect("is well-formatted yaml");
+                assert!(test_case.aggregate_verify())
```
ralexstokes, line 297: a minor thing but if you want to go the route of attaching the verification logic inside some "verify" method of some `FooTestIO` struct, i'd consider just calling the method `verify` or `execute` or something

then you can more easily copy/paste this code to use in multiple places -- may save you some work here

ralexstokes: if you moved to something like `execute`, i'd actually change the type signature to `() -> ()` so that you can drop the `assert!` inline here.

then up above in each test case, you can `assert!` accordingly

i think it is a weak idiom to get some computed result and `assert_eq!` against the expected/provided result so it may seem a bit more at home to a rust dev seeing this code if you follow the same pattern

tests/bls_tests.rs
pkopparla: It is this: [sign_case_zero_privkey](https://media.githubusercontent.com/media/ethereum/consensus-spec-tests/master/tests/general/phase0/bls/sign/small/sign_case_zero_privkey/data.yaml)

tests/bls_tests.rs
pkopparla: > one more thing i'd consider is moving the data parsing into `serde`
> 
> there are field attributes you can use to specify the desired transformation and that should keep the function bodies smaller and more focused by moving the parsing code elsewhere...
> 
> to give an example, `SigningTestIO` would look like:
> 
> ```rust
> #[derive(Debug, Serialize, Deserialize)]
> struct SigningInput {
>     #[serde(deserialize_with = "try_decode_hex_with_prefix")]
>     privkey: Option<SecretKey>,
>     #[serde(deserialize_with = "decode_hex_with_prefix")]
>     message: Vec<u8>,
> }
> 
> #[derive(Debug, Serialize, Deserialize)]
> struct SigningTestIO {
>     input: SigningInput,
>     #[serde(deserialize_with = "decode_hex_with_prefix")]
>     output: Vec<u8>,
> }
> ```
> 
> where you need to add a "fallible" version of `decode_hex_with_prefix` that would return an `None: Option<SecretKey>` if the key fails to parse.
> 
> you would also need to adapt the function signature of `decode_hex_with_prefix` to match the expected signature, here is an example: https://stackoverflow.com/a/46755370 (under the first header with their fn `from_hex`).
> 
> and this would be adjusted for each test type...

On this point, I'm struggling a little to understand how to make the Deserializer sufficiently generic to generate options for `PublicKey`, `Signature` and `SecretKey`.  In the stackoverflow example, the return type is set to `Result<u64>`. I'm trying something like this:
```
fn decode_hex_with_prefix<T: AsRef<[u8]> + Debug>(data: T) -> Vec<u8> {
    let data = data.as_ref();
    let data= if data.starts_with(b"0x") {
        &data[2..] } else {
            &data
        };
        hex::decode(data).expect("is not well-formed hex")
    }
fn try_decode_hex_with_prefix<T: AsRef<[u8]> + Debug>(data: T) -> Option<Vec<u8>> {
    match SecretKey::from_bytes(&decode_hex_with_prefix(data)) {
        Ok(value) => Some(value),
        Err(_) => None,
    }
}
```
but this will only return Option<SecretKey>. How to generalize this?

tests/bls_tests.rs
```diff
@@ -73,37 +59,26 @@ impl AggregatingTestIO {
...
 #[derive(Debug, Serialize, Deserialize)]
-
 struct AggVerifyTestIO {
     input: AggVerifyInput,
     output: bool,
 }
-
 impl AggVerifyTestIO {
-    fn aggregate_verify(&self) -> bool {
+    fn verify(&self) -> () {
```
ralexstokes, line 81: can drop the `()`:

```suggestion
    fn verify(&self) {
```

and other places here

tests/bls_tests.rs
```diff
@@ -131,34 +100,26 @@ impl AggVerifyTestIO {
...
-                Err(_) => {
-                    // handling for the zero signature case which raises a blst bad encoding error
-                    if self.output == false {
-                        return true;
-                    } else {
-                        return false;
-                    }
-                }
+                // handling for the zero signature case which raises a blst bad encoding error
+                Err(_) => return assert_eq!(self.output, false),
```
ralexstokes, line 104: i think you want this:

```suggestion
                Err(_) => return self.output == false,
```

ralexstokes: and this starts to get a little opaque but `return self.output == false` should be the same as `return !self.output` right?

pkopparla: Yes, but `==` returns `bool` but we want `()` which is returned by `assert_eq!`.

pkopparla: Ah, I'll change it to that then.

pkopparla:
Is this ok to merge? Any other changes needed?

tests/README.md
ralexstokes: did you want to keep this in? 

i'd suggest just removing it so we don't have the additional maintenance burden

tests/bls_tests.rs
```diff
@@ -0,0 +1,262 @@
...
+    message: String,
+}
+#[derive(Debug, Serialize, Deserialize)]
+struct SigningTestIO {
+    input: SigningInput,
+    output: String,
+}
+impl SigningTestIO {
+    fn verify(&self) {
+        // let input = &self.input;
```
ralexstokes, line 32: remove comment?

tests/bls_tests.rs
```diff
@@ -0,0 +1,262 @@
...
+        let input_signatures: Vec<Signature> = self
+            .input
+            .iter()
+            .map(|x| Signature::from_bytes(&decode_hex_with_prefix(&x)).unwrap())
+            .collect();
+        let aggregate = match aggregate(&input_signatures) {
+            Ok(agg) => agg,
+            // handling for zero sized input and output
+            Err(_) => {
+                assert_eq!(self.output, "~");
```
ralexstokes, line 65: can you show me which test case has the string `"~"`? i think this is some artifact of the yaml en/decoding

tests/bls_tests.rs
```diff
@@ -0,0 +1,262 @@
...
+        glob("consensus-spec-tests/tests/general/phase0/bls/fast_aggregate_verify/small/**/*.yaml")
+            .expect("Failed to read glob pattern");
+    for entry in entries {
+        let path = entry.unwrap();
+        let file = File::open(path).expect("File does not exist");
+        let test_case: FastAggVerifyTestIO =
+            serde_yaml::from_reader(file).expect("Is not well-formatted yaml");
+        test_case.verify()
+    }
+}
```
ralexstokes, line 213: if we have this general pattern across the test repos, what do you think about making a generic driver function?

something like:

```rust
fn execute_test_case<T>(path_glob: TheGlobType) {
    let entries =
        glob(path_glob)
            .expect("Failed to read glob pattern");
    for entry in entries {
        let path = entry.unwrap();
        let file = File::open(path).expect("File does not exist");
        let test_case: T =
            serde_yaml::from_reader(file).expect("Is not well-formatted yaml");
        test_case.verify()
    }
}
```

it may make more sense to write another test format before deciding on the right abstraction here but we should be able to abstract over the general "load this test and then run it" mechanism to avoid a lot of code duplication

tests/bls_tests.rs
```diff
@@ -0,0 +1,262 @@
...
+            match PublicKey::from_bytes(&decode_hex_with_prefix(&self.input.pubkey)) {
+                Ok(pk) => pk,
+                // error handling for infinity pub key
+                Err(_) => {
+                    assert!(!self.output);
+                    return;
+                }
+            };
+        let message_vec: Vec<u8> = decode_hex_with_prefix(&self.input.message);
+        let message: &[u8] = &message_vec;
```
ralexstokes, line 187: i'd drop the extra name `message_vec` here....

```rust
let message = decode_hex_with_prefix(...);
// down below
... = signature.verify(&pubkey, &message);
```

ralexstokes (commented):
great updates!

in general, i'd prefer one blank line between separate items to help with visual parsing

also did you want to add the serde feature to parse the hex string as bytes?

and i would suggest going through and removing as many type annotations as you can. it looks like you went through and annotated most (?) of the declarations but the rust compiler can resolve most of them and it just adds more visual noise to the code

tests/bls_tests.rs
pkopparla: It's this: https://media.githubusercontent.com/media/ethereum/consensus-spec-tests/master/tests/general/phase0/bls/aggregate/small/aggregate_na_signatures/data.yaml

pkopparla:
I've put the `execute_test_cases` and `verify` in a trait and reused it across all the IO structs. That was very satisfying :)
The documentation is minimized and the serde read-in feature is being used now. There's still one pattern that repeats over and over, where we read-in to the `crypto.rs` structs and check for errors:
```rust
let pubkeys_result: Result<Vec<PublicKey>, _> = self
            .input
            .pubkeys
            .iter()
            .map(|x| PublicKey::from_bytes(&x))
            .collect();
let pubkeys = match pubkeys_result {
            Ok(pk) => pk,
            // error handling for infinity pub key
            Err(_) => {
                assert!(!self.output);
                return;
            }
```
Could we try to reduce this or is it fine as is?

ralexstokes:
hi @pkopparla !

i made some code cleanups from your current PR here:

https://github.com/ralexstokes/ethereum_consensus/commit/45232e1dbb0ccbafca82934eaea20c4e683deaa2

i'd suggest moving them over from that commit to this PR :)

the main things were:

1) avoid type annotations when we can
2) cleaning up your trait design to just use `Self`, rather than the self-referential thing you had going on

tests/bls_tests.rs
```diff
@@ -0,0 +1,299 @@
...
+trait TestDriver<T: TestDriver<T> + for<'de> serde::Deserialize<'de>> {
+    fn verify(&self) {
+        ()
+    }
+
+    fn execute_test_cases(path_glob: &str) {
+        let entries = glob(path_glob).expect("Failed to read glob pattern");
+        for entry in entries {
+            let path = entry.unwrap();
+            println!("{:?}", path);
```
ralexstokes, line 258:
```suggestion
```

pkopparla:
Thanks for the comments @ralexstokes! I didn't know that `Self` trick. 
Does this code need anymore changes or can I move onto another category of tests? What would be next in order of priority?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "consensus-spec-tests"]
+	path = consensus-spec-tests
+	url = https://github.com/ethereum/consensus-spec-tests.git
```
ralexstokes, line 3: nitpick, but: do we need this leading whitespace?

tests/bls_tests.rs
```diff
@@ -0,0 +1,297 @@
...
+    } else {
+        data
+    };
+    //null character case, empty array
+    if data == [126] {
+        return Ok(vec![]);
+    }
+    hex::decode(data).map_err(D::Error::custom)
+}
+fn decode_hexvec_with_prefix<'a, D>(input: D) -> Result<Vec<Vec<u8>>, D::Error>
```
ralexstokes, line 30: nitpick, but i'd suggest an extra `_` here:
```suggestion
fn decode_hex_vec_with_prefix<'a, D>(input: D) -> Result<Vec<Vec<u8>>, D::Error>
```

tests/bls_tests.rs
```diff
@@ -0,0 +1,309 @@
...
+    aggregate, aggregate_verify, fast_aggregate_verify, PublicKey, SecretKey, Signature,
+};
+use glob::glob;
+use hex;
+use serde::{de::Error, Deserialize, Deserializer, Serialize};
+use serde_yaml;
+use std::fmt::Debug;
+use std::fs::File;
+
+// type TestResult = Result<(), Box<dyn Error>>;
```
ralexstokes, line 13:
```suggestion
```

ralexstokes (approved):
one minor cleanup, otherwise i think we move ahead w/ merge!

tests/bls_tests.rs
pkopparla: Done!

PR #18, opened 2021-11-10
Add key validation

dnkolegov:


src/crypto.rs
```diff
@@ -5,7 +5,9 @@ use ssz_rs::prelude::*;
 use std::fmt;
 use thiserror::Error;
 
-const BLS_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
+pub const BLS_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
```
ralexstokes, line 8:
```suggestion
const BLS_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
```

why is the `BLS_DST` now `pub`?

src/crypto.rs
```diff
@@ -98,6 +101,40 @@ impl SecretKey {
     }
 }
 
+impl Sized for SecretKey {
```
ralexstokes, line 104: why do we need these `impl`s for `SecretKey`?

ralexstokes, line 104: why do we need these `impl`s for `SecretKey`?

src/crypto.rs
```diff
@@ -135,12 +172,11 @@ impl PublicKey {
         self.0.validate().is_ok()
     }
 
-    pub fn from_bytes(encoding: &[u8]) -> Result<Self, Error> {
-        let inner = blst_core::PublicKey::from_bytes(encoding).map_err(BLSTError::from)?;
```
ralexstokes, line 139: i'd suggest something with fewer changes here:

```rust
        blst_core::PublicKey::key_validate(encoding).map(Self).map_err(BLSTError::from);
```

src/crypto.rs
```diff
@@ -5,7 +5,9 @@ use ssz_rs::prelude::*;
 use std::fmt;
 use thiserror::Error;
 
-const BLS_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
+pub const BLS_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
+pub const BLS_PUBLIC_KEY_BYTES_LEN: usize = 48;
+pub const BLS_SECRET_KEY_BYTES_LEN: usize = 32;
```
ralexstokes, line 10: do these need to be `pub`?

ralexstokes (commented):
looks good! i really like the testing you added that goes along w/ these changes.

i'm not sure why you added the `ssz::{Serialize, Deserialize}` impls for `SecretKey`; i think we can drop them as we don't currently have a need for those routines for secret keys (whereas we already serialize pub keys and sigs in phase 0, etc.). i think we can just keep the `blst_core::SecretKey::from_bytes` approach we originally had

and regarding https://github.com/ralexstokes/ethereum_consensus/pull/17#issuecomment-964798245, i'd suggest swapping where the implementation lives:

right now, we have `PublicKey::from_bytes` that calls to `Self::deserialize` and then has to cover up any `blst` error, along with introducing a new error variant `DeserializeError`. instead we can just let the `blst` error do the heavy lifting here (which it is already doing) so that `PublicKey::from_bytes` has the call to `key_validate` which then exposes the `blst` error, if there is one.

the `ssz::Deserialize` implementation can then call `Self::from_bytes` and we can do any error conversion we need to do there.

the rationale is that you may never touch the `ssz` machinery if you were using a `PublicKey` and this strategy I'm outlining only exposes `blst` concepts to the user until they opt-in to using some `ssz` trait.

if you agree then you can swap them and remove the `DeserializeError` variant as well bc then i think we don't need it

src/crypto.rs
dnkolegov: fixed.

src/crypto.rs
dnkolegov: fixed

src/crypto.rs
dnkolegov: fixed

src/crypto.rs
dnkolegov: fixed

src/crypto.rs
```diff
@@ -353,6 +316,11 @@ mod tests {
     use crate::crypto::{aggregate, aggregate_verify, fast_aggregate_verify};
     use rand::prelude::*;
 
+    const INFINITY_COMPRESSED_PUBLIC_KEY: [u8; BLS_PUBLIC_KEY_BYTES_LEN] = [
+        0xc0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    ];
+
```
ralexstokes, line 323: very nice test case

src/crypto.rs
```diff
@@ -462,10 +430,10 @@ mod tests {
...
 
     #[test]
-    #[should_panic(expected = "long public key")]
-    fn long_public_key() {
-        let z = [0u8; BLS_PUBLIC_KEY_BYTES_LEN + 1];
-        PublicKey::from_bytes(&z).expect("can make a long public key");
+    #[should_panic(expected = "infinity-based public key")]
+    fn infinity_as_public_key() {
+        PublicKey::from_bytes(&INFINITY_COMPRESSED_PUBLIC_KEY)
+            .expect(" can make an infinity-based public key");
```
ralexstokes, line 436: i still like the "input too long" test case, i can add it back

ralexstokes (commented):
nice updates! thanks, im going to make a few small syntactical changes and then merge it in :)

ralexstokes:
CI failing due to temporarily unused fns supporting the phase 0 spec but the CI does pass locally so proceeding with merge

PR #20, opened 2021-11-13
Add state transition context

ralexstokes:
We will need some sort of `Context` for a variety of state transition optimization we will want, and this also is a nice place to pass around some configuration data that is a bit unwieldy to pass around in other ways (e.g. thru compile-time constants)


ralexstokes:
i'll copy the commit message that i think summarizes the need here:

add Context type to state transition 

- lay ground work for future optimizations
- pass around "compile-time" constants that are prohibitive to use
  as const generics

as the situation around compile-time constant inference improves, we can shrink the `Context` over time... for now this should be sufficient

PR #22, opened 2021-11-15
Phase 0: Beacon state accessors

dnkolegov:


src/phase0/beacon_state.rs
```diff
@@ -61,3 +70,170 @@ pub struct BeaconState<
...
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    _context: &Context,
+) -> Epoch {
+    compute_epoch_at_slot(state.slot, _context)
```
ralexstokes, line 96: i only use the leading `_` in variable names when the variable is unused -- here it is passed to the inner function so no need for it

src/phase0/state_transition/mod.rs
```diff
@@ -38,7 +38,7 @@ pub enum Error {
     InsufficientValidators,
 }
 
-pub fn is_active_validator(validator: Validator, epoch: Epoch) -> bool {
+pub fn is_active_validator(validator: &Validator, epoch: Epoch) -> bool {
```
ralexstokes, line 41: yeah, i had noticed we will want to pass most of the bigger types by ref, nice catch here

src/phase0/beacon_state.rs
ralexstokes: and this goes for the rest of the uses below

ralexstokes (commented):
great start!

i don't think this is clear but i was thinking the "top-level" files outside of the `state_transition` module would just be the types and everything supporting the logic of the state transition would go inside the `state_transition` module.

i don't know exactly but i can foresee needing to add lots of additional structure to support a more optimized beacon state and in that case i'd think it best to put that logic in `crate::phase0::beacon_state` (as it belongs to the type)

i'd suggest moving the "accessors" you have into some (sub)module of `crate::phase0::state_transition` -- and possibly into a new file but this stuff could just live in `mod.rs`

src/phase0/beacon_state.rs
```diff
@@ -61,3 +70,170 @@ pub struct BeaconState<
     pub current_justified_checkpoint: Checkpoint,
     pub finalized_checkpoint: Checkpoint,
 }
+
+pub fn get_current_epoch<
```
ralexstokes, line 74: and if you consolidate/move the accessors into the `state_transition` module you won't duplicate the `get_current_epoch` function

src/phase0/beacon_state.rs
dnkolegov: fixed

src/phase0/beacon_state.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
```diff
@@ -631,3 +631,104 @@ pub fn get_active_validator_indices<
...
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    epoch: Epoch,
+    domain_type: DomainType,
+) -> Bytes32 {
+    let epoch = epoch + (EPOCHS_PER_HISTORICAL_VECTOR as u64 - MIN_SEED_LOOKAHEAD) - 1;
+    let mix = get_randao_mix(state, epoch);
+    let mut input = [0u8; 44];
+    input[..4].copy_from_slice(&domain_type.as_bytes());
+    input[4..].copy_from_slice(&epoch.to_le_bytes());
```
ralexstokes, line 695: i think `copy_from_slice` will panic if the `src` and `dst` are not the same length so you want to properly size the `input[4..8]` here

dnkolegov: fixed

src/phase0/state_transition/mod.rs
```diff
@@ -485,3 +491,358 @@ pub fn compute_fork_data_root(
...
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    indices: Vec<ValidatorIndex>,
+    context: &Context,
+) -> Result<Gwei, Error> {
+    let total_balance = indices
+        .into_iter()
+        .try_fold(0usize, |acc, i| {
+            acc.checked_add(state.validators[i].effective_balance as usize)
+        })
```
ralexstokes, line 845: is there a reason we can't stick with `u64` arithmetic here?

`try_fold` with `0u64` starting accumulator?

dnkolegov: sure

src/phase0/state_transition/mod.rs
```diff
@@ -761,20 +761,21 @@ pub fn get_beacon_committee<
     slot: Slot,
     index: CommitteeIndex,
     context: &Context,
-) -> Result<&[ValidatorIndex], Error> {
+) -> Result<Vec<ValidatorIndex>, Error> {
     let epoch = compute_epoch_at_slot(slot, context);
-    let committees_per_slot = get_committee_count_per_slot(state, epoch);
+    let committees_per_slot = get_committee_count_per_slot(&state, epoch, context);
```
ralexstokes, line 766: we will want to pass the state by reference most places

so the type signature can change to `&BeaconState` in the fn defn

and we don't need to keep taking `&` inside the fn body

src/phase0/state_transition/mod.rs
```diff
@@ -761,20 +761,21 @@ pub fn get_beacon_committee<
     slot: Slot,
     index: CommitteeIndex,
     context: &Context,
-) -> Result<&[ValidatorIndex], Error> {
+) -> Result<Vec<ValidatorIndex>, Error> {
```
ralexstokes, line 764: what was wrong w/ the slice here?

src/phase0/state_transition/mod.rs
```diff
@@ -836,13 +833,117 @@ pub fn get_total_balance<
...
+        SLOTS_PER_HISTORICAL_ROOT,
+        HISTORICAL_ROOTS_LIMIT,
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    attestation: &mut Attestation<MAX_VALIDATORS_PER_COMMITTEE>,
```
ralexstokes, line 892: why is this `mut`?

we should not mutate the `attestation` to get the matching `IndexedAttestation`

src/phase0/state_transition/mod.rs
```diff
@@ -836,13 +833,117 @@ pub fn get_total_balance<
...
+        HISTORICAL_ROOTS_LIMIT,
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    data: &AttestationData,
+    bits: &mut Bitlist<MAX_VALIDATORS_PER_COMMITTEE>,
```
ralexstokes, line 928: and similarly, we should not mutate `bits` -- does something break if this is not `mut`?

src/phase0/state_transition/mod.rs
```diff
@@ -836,13 +833,117 @@ pub fn get_total_balance<
...
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    attestation: &mut Attestation<MAX_VALIDATORS_PER_COMMITTEE>,
+    context: &Context,
+) -> Result<IndexedAttestation<MAX_VALIDATORS_PER_COMMITTEE>, Error> {
+    let bits = &mut attestation.aggregation_bits;
+    let attesting_indices = get_attesting_indices(state, &attestation.data, bits, context)?;
+    attesting_indices.to_vec().sort();
```
ralexstokes, line 897: this is already a `Vec` and `sorted`? i think you can delete this line

src/phase0/state_transition/mod.rs
```diff
@@ -2,13 +2,14 @@ mod block_processing;
 mod context;
 mod epoch_processing;
 
+use std::borrow::Borrow;
```
ralexstokes, line 5: do we need Borrow here?

src/phase0/state_transition/mod.rs
```diff
@@ -18,6 +19,7 @@ pub use context::Context;
 use ssz_rs::prelude::*;
 use std::cmp;
 use std::collections::HashSet;
+use std::ops::DerefMut;
```
ralexstokes, line 22: do we need this trait in scope?

ralexstokes (commented):
nice work on the further progress!

left some comments below

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: accidentally

src/phase0/state_transition/mod.rs
dnkolegov: we return a reference to a local variable

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: fixed

Cargo.toml
```diff
@@ -16,3 +16,4 @@ thiserror = "1.0.30"
 sha2 = "0.9.8"
 hex = "0.4.3"
 integer-sqrt = "0.1.5"
+itertools = "0.10.0"
```
ralexstokes, line 19:
```suggestion
itertools = "0.10.1"
```

?

src/phase0/state_transition/mod.rs
```diff
@@ -485,3 +556,467 @@ pub fn compute_fork_data_root(
...
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    attestation: &Attestation<MAX_VALIDATORS_PER_COMMITTEE>,
+    context: &Context,
+) -> Result<IndexedAttestation<MAX_VALIDATORS_PER_COMMITTEE>, Error> {
+    let bits = &attestation.aggregation_bits;
+    let mut attesting_indices = get_attesting_indices(state, &attestation.data, bits, context)?;
+    attesting_indices.sort_unstable();
```
ralexstokes, line 968: this is a detail but the spec used python's `sorted` which is a *stable* sort:

> The built-in sorted() function is guaranteed to be stable. 

https://docs.python.org/3/library/functions.html#sorted

src/phase0/state_transition/mod.rs
```diff
@@ -485,3 +556,467 @@ pub fn compute_fork_data_root(
...
+    epoch: Epoch,
+) -> Vec<ValidatorIndex> {
+    let mut active: Vec<ValidatorIndex> = Vec::with_capacity(state.validators.len());
+
+    for (i, v) in state.validators.iter().enumerate() {
+        if is_active_validator(v, epoch) {
+            active.push(i)
+        }
+    }
+    active
```
ralexstokes, line 704: don't need to necessarily change this now but i will point out this could be a bit more compactly expressed with `filter_map` on `state.validators.iter().enumerate()`

src/phase0/state_transition/mod.rs
```diff
@@ -485,3 +556,467 @@ pub fn compute_fork_data_root(
...
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    data: &AttestationData,
+    bits: &Bitlist<MAX_VALIDATORS_PER_COMMITTEE>,
+    context: &Context,
+) -> Result<Vec<ValidatorIndex>, Error> {
```
ralexstokes, line 1001: we should return a `HashSet<ValidatorIndex>` here... other places in the spec fit more nicely with this being a set type

and then we don't need the `unique` call which i believe is an addition from `itertools`

src/phase0/state_transition/mod.rs
ralexstokes: so to handle this i would:

1) update `get_attesting_indices` to return `HashSet`
2) still make `attesting_indices` as `List::from_iter` and then call `attesting_indices.sort()`
3) supply this value to the `IndexedAttestation`

this API is changing a bit in a PR i'm working on for `ssz_rs` but i will come back and fix that when it lands

src/phase0/state_transition/mod.rs
```diff
@@ -106,16 +141,50 @@ pub fn is_valid_indexed_attestation<
...
+                    Err(Error::InvalidOperation(
+                        InvalidOperation::IndexedAttestation(
+                            InvalidIndexedAttestation::AttestingIndicesNotSorted,
+                        ),
+                    ))
+                }
+            })?;
+        Ok(())
+    };
+    check_sorted(attesting_indices.as_slice())?;
```
ralexstokes, line 170: i think this can be done more simply and we can drop the dependency on `itertools`:

something like:

```rust
let is_sorted = attesting_indices.windows(2).map(|pair| { let a = &pair[0]; let b = &pair[1]; a < b }).all(|x| x);
if !is_sorted {
    return Err(...)
}
```

src/phase0/state_transition/mod.rs
```diff
@@ -106,16 +141,50 @@ pub fn is_valid_indexed_attestation<
...
+        return Err(Error::InvalidOperation(
+            InvalidOperation::IndexedAttestation(InvalidIndexedAttestation::AttestingIndicesEmpty),
+        ));
     }
-    let indices: HashSet<usize> =
-        HashSet::from_iter(indexed_attestation.attesting_indices.iter().cloned());
+    let check_sorted = |list: &[ValidatorIndex]| -> Result<(), Error> {
+        list.iter()
+            .tuple_windows()
+            .enumerate()
```
ralexstokes, line 156: note here you called `enumerate` but then it looks like you just discard the index in `|(_, (x, y))|` below

Cargo.toml
ralexstokes: also see below, i think we can hold off on adding `iterools` for now

ralexstokes (commented):
nice updates! i left some suggestions to simplify below

src/phase0/state_transition/mod.rs
dnkolegov: ok

src/phase0/state_transition/mod.rs
dnkolegov: ok

src/phase0/state_transition/mod.rs
dnkolegov: ok

src/phase0/state_transition/mod.rs
dnkolegov: ok

src/phase0/state_transition/mod.rs
```diff
@@ -893,7 +891,7 @@ pub fn get_total_balance<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    indices: &[ValidatorIndex],
+    indices: HashSet<ValidatorIndex>,
```
ralexstokes, line 894: ah nice catch!

src/phase0/state_transition/mod.rs
```diff
@@ -1004,13 +1000,14 @@ pub fn get_attesting_indices<
         )));
     }
 
-    let mut indices = Vec::with_capacity(bits.capacity());
+    //let mut indices = Vec::with_capacity(bits.capacity());
```
ralexstokes, line 1003:
```suggestion
```

src/phase0/state_transition/mod.rs
```diff
@@ -485,3 +554,459 @@ pub fn compute_fork_data_root(
...
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    epoch: Epoch,
+) -> Vec<ValidatorIndex> {
+    let mut active: Vec<ValidatorIndex> = Vec::with_capacity(state.validators.len());
```
ralexstokes, line 695:
```suggestion
    let mut active = Vec::with_capacity(state.validators.len());
```

PR #23, opened 2021-12-02
Add beacon state mutators

dnkolegov:


src/phase0/presets/mod.rs
```diff
@@ -36,4 +36,7 @@ pub struct Preset {
     pub max_attestations: usize,
     pub max_deposits: usize,
     pub max_voluntary_exits: usize,
+    pub min_validator_withdrawability_delay: Epoch,
+    pub churn_limit_quatient: u64,
```
ralexstokes, line 40: spelling:

```suggestion
    pub churn_limit_quotient: u64,
```

src/phase0/state_transition/mod.rs
```diff
@@ -842,9 +840,6 @@ pub fn get_beacon_proposer_index<
     const EPOCHS_PER_SLASHINGS_VECTOR: usize,
     const MAX_VALIDATORS_PER_COMMITTEE: usize,
     const PENDING_ATTESTATIONS_BOUND: usize,
-    const MAX_COMMITTEES_PER_SLOT: u64,
```
ralexstokes, line 845: so these were just extra generics? i suppose this will become clear as we use these functions elsewhere but something to keep an eye on!

src/phase0/state_transition/mod.rs
```diff
@@ -1009,3 +1004,172 @@ pub fn get_attesting_indices<
...
+    index: ValidatorIndex,
+    context: &Context,
+) {
+    if state.validators[index].exit_epoch != FAR_FUTURE_EPOCH {
+        return;
+    }
+
+    let mut exit_epochs: Vec<Epoch> = state
+        .validators
+        .as_slice()
```
ralexstokes, line 1093: this may not be possible right now but the `as_slice` call should be unnecessary -- i may need to merge in the changes from https://github.com/ralexstokes/ssz_rs/pull/3 though...

src/phase0/state_transition/mod.rs
```diff
@@ -1009,3 +1004,172 @@ pub fn get_attesting_indices<
...
+    exit_epochs.push(compute_activation_exit_epoch(
+        get_current_epoch(state, context),
+        context,
+    ));
+
+    let mut exit_queue_epoch = *exit_epochs.iter().max().unwrap();
+
+    let exit_queue_churn = state
+        .validators
+        .as_slice()
```
ralexstokes, line 1108: ditto on `as_slice` here -- can we not get to `iter` without it?

src/phase0/state_transition/mod.rs
```diff
@@ -1009,3 +1004,172 @@ pub fn get_attesting_indices<
...
+        .filter(|v| v.exit_epoch == exit_queue_epoch)
+        .count();
+
+    if exit_queue_churn >= get_validator_churn_limit(state, context) as usize {
+        exit_queue_epoch += 1
+    }
+
+    state.validators[index].exit_epoch = exit_queue_epoch;
+    state.validators[index].withdrawable_epoch =
+        state.validators[index].exit_epoch + context.min_validator_withdrawability_delay as u64;
```
ralexstokes, line 1119: what is the `as u64` doing here?

src/phase0/state_transition/mod.rs
```diff
@@ -1009,3 +1004,172 @@ pub fn get_attesting_indices<
...
+    slashed_index: ValidatorIndex,
+    whistleblower_index: Option<ValidatorIndex>,
+    context: &Context,
+) -> Result<(), Error> {
+    let epoch = get_current_epoch(state, context);
+    initiate_validator_exit(state, slashed_index, context);
+    state.validators[slashed_index].slashed = true;
+    state.validators[slashed_index].withdrawable_epoch = u64::max(
+        state.validators[slashed_index].withdrawable_epoch,
+        epoch + context.epochs_per_slashings_vector as u64,
```
ralexstokes, line 1151: same question on the `as u64` here

src/phase0/state_transition/mod.rs
```diff
@@ -1009,3 +1004,172 @@ pub fn get_attesting_indices<
...
+    );
+    state.slashings[epoch as usize % EPOCHS_PER_SLASHINGS_VECTOR as usize] +=
```
ralexstokes, line 1153: imo a bit more readable to move the expression inside the `[..]` to the prior line as a `slashings_index` or something like that

src/phase0/state_transition/mod.rs
```diff
@@ -727,8 +725,8 @@ pub fn get_validator_churn_limit<
     let active_validator_indices =
         get_active_validator_indices(state, get_current_epoch(state, context));
     u64::max(
-        MIN_PER_EPOCH_CHURN_LIMIT as u64,
-        active_validator_indices.len() as u64 / CHURN_LIMIT_QUOTIENT as u64,
+        context.min_per_epoch_churn_limit as u64,
+        active_validator_indices.len() as u64 / context.churn_limit_quatient as u64,
```
ralexstokes, line 729: the values from the `context` here are already `u64` right? so no need to cast?

ralexstokes (commented):
excellent work on the mutators! i left some minor feedback/comments above.

there is only one other thing to change and it is how you added the `config` values.

my intention is to mirror the specs layout so that there is a `config` type separate from a `preset` type. the presets are things that don't change as much from testnet to testnet and we can even go ahead and provide them statically (like we do for `mainnet` and will do for `minimal` once the overall architecture stabilizes). 

the config values are much more "dynamic" in this sense and can change from testnet to testnet and usually these values are even loaded from a YAML file provided at runtime.

so a few things i see that will differ from what we have now:

1. remove config values from the preset type/file.
2. we do want them in the `context`, but the context should load them from a `Config` type.
3. we have the beginning of a `config` module under `phase0::configs` and this is where a `Config` type should live.

i'd suggest following the pattern with presets where we have a `Config` type that we can hardcode for mainnet, as we know the small handful of values. this type provides a hook for later when we will inevitably want to load a bespoke set of configuration from some YAML file.

this make sense? i would even consider making a separate PR w/ the config changes that bubble up into the `Context` type and then we can merge that into main and rebase this PR on top of that. then this PR can just be the few changes for the mutations you already have

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/mod.rs
dnkolegov: fixed

src/phase0/state_transition/context.rs
```diff
@@ -37,9 +38,18 @@ pub struct Context {
     pub max_attestations: usize,
     pub max_deposits: usize,
     pub max_voluntary_exits: usize,
+
```
ralexstokes, line 41:
```suggestion
```

?

src/phase0/state_transition/context.rs
```diff
@@ -77,9 +87,19 @@ impl Context {
             max_attestations: preset.max_attestations,
             max_deposits: preset.max_deposits,
             max_voluntary_exits: preset.max_voluntary_exits,
-            churn_limit_quatient: preset.churn_limit_quatient,
-            min_validator_withdrawability_delay: preset.min_validator_withdrawability_delay,
-            min_per_epoch_churn_limit: preset.min_per_epoch_churn_limit,
+
```
ralexstokes, line 90: drop this whitespace?

src/phase0/state_transition/mod.rs
```diff
@@ -1150,8 +1148,8 @@ pub fn slash_validator<
         state.validators[slashed_index].withdrawable_epoch,
         epoch + context.epochs_per_slashings_vector as u64,
     );
-    state.slashings[epoch as usize % EPOCHS_PER_SLASHINGS_VECTOR as usize] +=
-        state.validators[slashed_index].effective_balance;
+    let slashing_index = epoch as usize % EPOCHS_PER_SLASHINGS_VECTOR;
+    state.slashings[slashing_index] += state.validators[slashed_index].effective_balance;
```
ralexstokes, line 1152:
```suggestion
    let slashings_index = epoch as usize % EPOCHS_PER_SLASHINGS_VECTOR;
    state.slashings[slashings_index] += state.validators[slashed_index].effective_balance;
```

just to make it clear this is the index into `slashings` and not related to the slashed validator

ralexstokes (commented):
great updates!

what do you think about making a `Config` type that mirrors the pattern with the `Preset`?

this type will be a target for a `load_from_yaml` method we will have in the future (as configs across networks change regularly) and it is also imo a bit easier to pass around one struct, rather than N constants defined elsewhere.

in this case the `Context::with_preset` function would change, perhaps to a method like `from(preset: &Preset, config: &Config) -> Self`

either way, the code in `phase0::state_transition::context` should be generic to which preset/config so we should not import any kind of `mainnet` or `minimal` here...

also, i would import the new mutators into `phase0::presets::mainnet` to silence the clippy warnings for now and get the CI passing

src/phase0/state_transition/context.rs
dnkolegov: ok

src/phase0/state_transition/context.rs
dnkolegov: ok

src/phase0/state_transition/mod.rs
dnkolegov: ok

src/phase0/presets/mainnet.rs
```diff
@@ -95,7 +97,7 @@ pub const PRESET: Preset = Preset {
 };
 
 pub fn context() -> Context {
-    Context::with_preset(&PRESET)
+    Context::from(&PRESET, &CONFIG)
 }
```
ralexstokes, line 101: following the symmetry here, there should be a `phase0::state_transition::context::mainnet` module and this should not live in the `preset`s

i'll make a note to fix this outside of this PR, it isn't super urgent

src/phase0/state_transition/context.rs
```diff
@@ -74,6 +86,18 @@ impl Context {
...
+            genesis_delay: config.genesis_delay,
+            seconds_per_slot: config.seconds_per_slot,
+            seconds_per_eth1_block: config.seconds_per_eth1_block,
+            min_validator_withdrawability_delay: config.min_validator_withdrawability_delay,
+            shard_committee_period: config.shard_committee_period,
+            eth1_follow_distance: config.eth1_follow_distance,
+            ejection_balance: config.ejection_balance,
+            churn_limit_quotient: config.churn_limit_quotient,
+            min_per_epoch_churn_limit: config.min_per_epoch_churn_limit,
+
```
ralexstokes, line 100:
```suggestion
```

PR #26, opened 2021-12-10
Slot processing

ralexstokes:


src/phase0/state_transition/slot_processing.rs
```diff
@@ -0,0 +1,77 @@
...
+        SLOTS_PER_HISTORICAL_ROOT,
+        HISTORICAL_ROOTS_LIMIT,
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    slot: &Slot,
```
ralexstokes, line 28: no need to pass this by ref here, can just use `Slot` directly

src/phase0/state_transition/mod.rs
```diff
@@ -1171,3 +1174,4 @@ pub fn slash_validator<
     );
     Ok(())
 }
+
```
ralexstokes, line 1177: extra whitespace?

src/phase0/state_transition/slot_processing.rs
```diff
@@ -0,0 +1,77 @@
...
+    let previous_state_root = state.hash_tree_root(merkleization_context)?;
+
+    state.state_roots[(state.slot % context.slots_per_historical_root as u64) as usize] = previous_state_root;
+    if state.latest_block_header.state_root.as_ref() == Bytes32::default().as_ref() {
+        state.latest_block_header.state_root = previous_state_root;
+    }
+    let previous_block_root = state.latest_block_header.hash_tree_root(merkleization_context)?;
+    state.block_roots[(state.slot % context.slots_per_historical_root as u64) as usize] = previous_block_root;
+    Ok(())
+}
```
ralexstokes, line 77: missing newline?

src/phase0/state_transition/mod.rs
```diff
@@ -44,6 +45,8 @@ pub enum Error {
     Overflow,
     #[error("{0}")]
     InvalidOperation(InvalidOperation),
+    #[error("overflow")]
```
ralexstokes, line 48:
```suggestion
    #[error("invalid slot provided")]
```

src/phase0/state_transition/slot_processing.rs
```diff
@@ -0,0 +1,77 @@
...
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    slot: &Slot,
+    context: &Context,
+) -> Result<(), Error> {
+    if state.slot < *slot {
+        return Err(InvalidSlot)
+    }
```
ralexstokes, line 33:
```suggestion
    if state.slot >= *slot {
        return Err(InvalidSlot)
    }
```

src/phase0/state_transition/slot_processing.rs
dnkolegov: ok

src/phase0/state_transition/mod.rs
dnkolegov: ok

src/phase0/state_transition/mod.rs
ralexstokes: want to extend this enum variant w/ data so the error is more descriptive?

```rust
#[error("an invalid transition to a past slot {requested} from slot {current}")]
TransitionToPreviousSlot { current: Slot, requested: Slot}
```

dnkolegov: ok

src/phase0/state_transition/slot_processing.rs
dnkolegov: good catch

src/phase0/state_transition/slot_processing.rs
```diff
@@ -1,8 +1,13 @@
 use crate::phase0::beacon_state::BeaconState;
-use crate::phase0::mainnet::Error::InvalidSlot;
-use crate::phase0::state_transition::{Context, Error};
+use crate::phase0::mainnet::Error::{InvalidEpoch, InvalidStateRoot, TransitionToPreviousSlot};
```
ralexstokes, line 2: style thing my preference here is to find these variants off of `Error`

and we want to use `Error` from the `state_transition` module

```suggestion
```

src/phase0/state_transition/slot_processing.rs
```diff
@@ -25,13 +30,16 @@ pub fn process_slots<
...
         PENDING_ATTESTATIONS_BOUND,
     >,
-    slot: &Slot,
+    slot: Slot,
     context: &Context,
 ) -> Result<(), Error> {
-    if state.slot < *slot {
-        return Err(InvalidSlot)
+    if state.slot >= slot {
+        return Err(TransitionToPreviousSlot {
```
ralexstokes, line 37:
```suggestion
        return Err(Error::TransitionToPreviousSlot {
```

src/phase0/state_transition/slot_processing.rs
```diff
@@ -67,11 +75,137 @@ pub fn process_slot<
...
+        .latest_block_header
+        .hash_tree_root(merkleization_context)?;
+    state.block_roots[(state.slot % context.slots_per_historical_root as u64) as usize] =
+        previous_block_root;
     Ok(())
-}
\ No newline at end of file
+}
+
+pub fn state_transition<
```
ralexstokes, line 91: another stylistic thing but i'd personally put a function like this into the `phase0::state_transition` top-level module, not these "sub-modules" for the individual sub-transitions

src/phase0/state_transition/slot_processing.rs
```diff
@@ -67,11 +75,137 @@ pub fn process_slot<
...
+    }
+    process_block(state, block, context)?;
+    if validate_result && block.state_root != state.hash_tree_root(merkleization_context)? {
+        return Err(InvalidStateRoot);
+    }
+
+    Ok(())
+}
+
+pub fn get_matching_source_attestations<
```
ralexstokes, line 143: these `get_matching_*` functions are part of the epoch processing so should live in that module

src/phase0/state_transition/slot_processing.rs
```diff
@@ -67,11 +75,137 @@ pub fn process_slot<
...
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    epoch: Epoch,
+    context: &Context,
+) -> Result<Vec<PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>>, Error> {
```
ralexstokes, line 165: something to consider if we start racking up very nested types is to define a type alias (next to `state_transition::Error`), something like:

```rust
type StateTransitionResult<T> = Result<T, Error>;
```

ralexstokes: then we can just write here:

```rust
StateTransitionResult<Vec<PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>>>
```

src/phase0/state_transition/slot_processing.rs
```diff
@@ -67,11 +75,137 @@ pub fn process_slot<
...
+    let previous_epoch = get_previous_epoch(state, context);
+    let current_epoch = get_current_epoch(state, context);
+
+    if previous_epoch != epoch && current_epoch != epoch {
+        return Err(InvalidEpoch {
+            previous: previous_epoch,
+            current: current_epoch,
+        });
```
ralexstokes, line 173: i'd update this error to include the "invalid epoch"


```rust
Err(Error::InvalidEpoch { requested: epoch, previous: previous_epoch, current: current_epoch }
```

src/phase0/state_transition/slot_processing.rs
```diff
@@ -67,11 +75,137 @@ pub fn process_slot<
...
+    }
+
+    if epoch == current_epoch {
+        return Ok(state.current_epoch_attestations.to_vec());
+    }
+    Ok(state.previous_epoch_attestations.to_vec())
```
ralexstokes, line 179: we should be able to just return a ref to the correct attestation list, rather than allocate an entirely new Vec for them...

i'd suggest something like this: https://gist.github.com/rust-play/bda5e26dcd644ec30b358966eb382731

src/phase0/state_transition/slot_processing.rs
```diff
@@ -67,11 +75,137 @@ pub fn process_slot<
...
+    epoch: Epoch,
+    context: &Context,
+) -> Result<Vec<PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>>, Error> {
+    let source_attestations = get_matching_source_attestations(state, epoch, context)?;
+    let block_root = get_block_root(state, epoch, context)?;
+    Ok(source_attestations
+        .into_iter()
+        .filter(|a| a.data.target.root.as_ref() == block_root.as_ref())
+        .collect())
+}
```
ralexstokes, line 211: i'd suggest dropping the intermediate allocations here, unless the ergonomics just gets radically better...

in this case, we can just return the existential type `impl Iterator`

```rust
get_matching_target_attestations<...>(state, epoch, context) -> Result<impl Iterator<Item=PendingAttestation<...>>, Errror>
```

and it may be worth adjusting the type signatures so they are more readable but that is something we can do later....


if we make this change then the last line would be:

```rust
Ok(source_attestations.into_iter().filter(|a| ...))
```

ralexstokes (commented):
excellent work!

i'd suggest putting the all of the epoch processing in future PRs, and keeping this one smaller with just the "per-slot" processing.

looks like the only thing missing for that is the `verify_block_signature` fn?

otherwise, i'd address the feedback/comments i left and then we should merge this in :)

src/phase0/state_transition/slot_processing.rs
dnkolegov: fixed

src/phase0/state_transition/slot_processing.rs
```diff
@@ -67,11 +75,137 @@ pub fn process_slot<
...
+    let validate_result = validate_result.unwrap_or(true);
+    let block = &signed_block.message;
+
+    process_slots(state, block.slot, context)?;
+    if validate_result {
+        verify_block_signature(state, signed_block, context)?;
+    }
+    process_block(state, block, context)?;
+    if validate_result && block.state_root != state.hash_tree_root(merkleization_context)? {
+        return Err(InvalidStateRoot);
```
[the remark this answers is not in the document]
dnkolegov: ok

src/phase0/state_transition/slot_processing.rs
dnkolegov: ok

src/phase0/state_transition/slot_processing.rs
dnkolegov: ok

src/phase0/state_transition/slot_processing.rs
dnkolegov: ok

src/phase0/state_transition/slot_processing.rs
dnkolegov: ok

src/phase0/state_transition/mod.rs
```diff
@@ -50,9 +50,13 @@ pub enum Error {
     #[error("invalid state root")]
     InvalidStateRoot,
     #[error(
-        "an invalid epoch is not matching to previous epoch {previous} and current epoch {current}"
+        "an invalid epoch {requested} is not matching to previous epoch {previous} and current epoch {current}"
```
ralexstokes, line 53: i'd suggest updating the language a bit:
```suggestion
        "the requested epoch {requested} is not in the required current epoch {current} or previous epoch {previous}"
```

dnkolegov: ok

src/phase0/state_transition/slot_processing.rs
dnkolegov: ok

src/phase0/state_transition/slot_processing.rs
dnkolegov: fixed

PR #30, opened 2021-12-23
feat: complete 'process_eth1_data_reset'

buchhlz2:


src/phase0/state_transition/epoch_processing.rs
```diff
@@ -202,8 +202,13 @@ pub fn process_eth1_data_reset<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<(), Error> {
+    let next_epoch = get_current_epoch(state, context) + 1;
+
+    if next_epoch % context.epochs_per_eth1_voting_period.into() == 0 {
```
ralexstokes, line 209: i don't see a reason to not strengthen the typing here on `epochs_per_eth1_voting_period` and then dropping the `into` call...

the diff will look something like this:

https://gist.github.com/ralexstokes/1430507703ed1cd00447113eefcb428f

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -202,8 +202,13 @@ pub fn process_eth1_data_reset<
...
+        state.eth1_data_votes = vec![];
```
ralexstokes, line 210: this fails if you look at the CI (or just run clippy or try to build locally) as the types don't line up!

the `BeaconState` is an SSZ object which means that each of its fields is an SSZ object, and in this case, `eth1_data_votes` is a `List` type.

so one thing we could do is just set an empty List: `List::new()`.

something marginally better is to keep the backing allocation (there is a `Vec` hidden behind some ssz machinery to make a `List`) and use a method like `clear`: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.clear

unfortunately, i have gone and hidden the mutating methods of `Vec` from our `List` type so that `List`s will be able to track their mutations to update the backing Merkle tree (which is currently WIP -- you can see initial progress on Vector here: https://github.com/ralexstokes/ssz_rs/pull/6

my current suggestion to address this issue is to use something like `clear` but we will need to extend `ssz_rs::List` with such a method (you can see i started w/ `push` and `pop`: https://github.com/ralexstokes/ssz_rs/blob/main/ssz_rs/src/list.rs#L194-L203)

ralexstokes (commented):
excellent work!

i left some comments on how to improve this PR further :)

src/phase0/state_transition/epoch_processing.rs
ralexstokes: this should do the trick: https://github.com/ralexstokes/ssz_rs/commit/68fa234abbfffa57b394e91ba9ce4344feff4305

and you may need to `cargo update -p ssz_rs` locally to pull the latest changes

ralexstokes: so it would just look like:

```rust
if ... {
    state.eth1_data_votes.clear();
}
```

buchhlz2:
I think I accidentally deleted the original PR (was trying to resolve some git issues) but it's now here: https://github.com/ralexstokes/ethereum_consensus/pull/31

PR #32, opened 2021-12-23
feat: complete body for fn 'process_slashings_reset'

buchhlz2:


src/phase0/state_transition/epoch_processing.rs
```diff
@@ -257,8 +257,11 @@ pub fn process_slashings_reset<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<(), Error> {
+    let next_epoch = get_current_epoch(state, context) + 1;
+
+    state.slashings[next_epoch as usize % context.epochs_per_slashings_vector] = Gwei::MIN;
```
ralexstokes, line 264: so this is correct but i do have to wonder about safety of using `MIN`.

*if* there was ever a case where we changed the `Gwei` type (e.g. it needed to become something other than the current `u64` for performance reasons), then it is possible that the meaning of `MIN` would change and this code would then break.

so imo it is a *bit* safer to just write:
```suggestion
    state.slashings[next_epoch as usize % context.epochs_per_slashings_vector] = 0;
```

that being said, i don't really see a reason we would ever need to change `Gwei` so i will let you decide which you prefer.

ralexstokes (approved):
nice! i'll go ahead and approve this and i *think* you should be able to merge it in at your option...

i left one thing to consider changing before merging it in below

src/phase0/state_transition/epoch_processing.rs
buchhlz2: My only rationale of this approach was to keep the type as `Gwei`. So, I think my only other options were to either create a variable like `let zero_gwei: Gwei = 0` or to add some method to `Gwei` so that I could say something like `Gwei::new(0)`.

Idk what the best practice is for using the type alias for readability purposes, so I'll let you decide lol...aka if just setting to zero is fine instead of actually using `Gwei` since they'e both a `u64`. Readability is nice but obviously not essential if it's not, technically, future-proof....

ralexstokes: yeah good thinking! 

the type of that term is already `Gwei`, if you write `Gwei::MIN` or if you write `0` -- the rust compiler would complain if you tried to put an instance of the wrong type into the field `state.slashings` (which has type `Vector<Gwei, ...>`) (e.g. try compiling `state.slashings = vec![]` or even `state.slashings = 23f64`). 

whether we write `Gwei::MIN` or `0`, it is telling the compiler the same thing in this instance.

the `let zero_gwei: Gwei = 0` would be redundant, e.g.:

```rust
let zero_gwei: Gwei = 0;
assert_eq!(zero_gwei, 0);
```

and we don't need anything like `Gwei::new` as the thinking behind using a type alias for this type was to skip as much ceremony as possible -- `Gwei` is simple enough that we can just leverage all of the functionality of Rust's `u64` type -- the only thing we are missing is the SSZ logic which has been provided in `ssz_rs`.


so in short, the type is still the same and then this one decision is sort of a matter of style -- but if you are e.g. doing math outside of rust, you likely use `0` where you can and not something like `min(Z)` where `Z` represents the integers... by analogy i don't see why we wouldn't just use `0` here :)

buchhlz2: Ahh got it, so Rust knows it's already `Gwei` -- okay yeah, that all makes sense...will merge with just `0`

Issue #29, opened 2021-12-22
Finish phase0 spec


ralexstokes:
Great progress has been made on the phase0 spec!

Here is a list of what is left:

- [x] process_epoch
  - [x]  process_justification_and_finalization https://github.com/ralexstokes/ethereum_consensus/pull/56
  - [x] process_rewards_and_penalties https://github.com/ralexstokes/ethereum_consensus/pull/46
  - [x] process_registry_updates: https://github.com/ralexstokes/ethereum_consensus/pull/42
  - [x] process_slashings: https://github.com/ralexstokes/ethereum_consensus/pull/39
  - [x] process_eth1_data_reset: https://github.com/ralexstokes/ethereum_consensus/pull/31
  - [x] process_effective_balance_updates https://github.com/ralexstokes/ethereum_consensus/pull/47
  - [x] process_slashings_reset: https://github.com/ralexstokes/ethereum_consensus/pull/32
  - [x] process_randao_mixes_reset:  https://github.com/ralexstokes/ethereum_consensus/pull/33
  - [x] process_historical_roots_update: https://github.com/ralexstokes/ethereum_consensus/pull/37
  - [x] process_participation_record_updates: https://github.com/ralexstokes/ethereum_consensus/pull/36
- [x] process_block
  - [x] process_block_header: https://github.com/ralexstokes/ethereum_consensus/pull/34
  - [x] process_randao: https://github.com/ralexstokes/ethereum_consensus/pull/40
  - [x] process_eth1_data: https://github.com/ralexstokes/ethereum_consensus/pull/41
  - [x] process_operations
    - [x] process_proposer_slashing: https://github.com/ralexstokes/ethereum_consensus/pull/45
    - [x] process_attester_slashing: https://github.com/ralexstokes/ethereum_consensus/pull/48
    - [x] process_attestation: https://github.com/ralexstokes/ethereum_consensus/pull/49
    - [x] process_deposit: https://github.com/ralexstokes/ethereum_consensus/pull/50
    - [x] process_voluntary_exit: https://github.com/ralexstokes/ethereum_consensus/pull/51

Each of these remaining bits are fairly self-contained and should be highly parallelizable.

ralexstokes:
just added skeletons for all the various sub-transitions in 68a652bbfc269ff142c7b36e97c78b6d7ae68eaa and e63b7e12050085d3e03c5b2157f610d435d200ce

buchhlz2:
[...]

PR #33, opened 2021-12-23
feat: add fn body for 'process_randao_mixes_reset'

buchhlz2:
Not sure if the usage of `clone()` is desired but had to get rid of the borrow...not sure if there's a more Rust way

Also lmk if I should keep the PR's specific to a fn, or if I should group them together in a PR so that I'm not spamming with PRs consisting of a few lines of code.

### DIFF
--- a/src/phase0/state_transition/epoch_processing.rs
+++ b/src/phase0/state_transition/epoch_processing.rs
@@ -1,7 +1,7 @@
 use crate::phase0::beacon_state::BeaconState;
 use crate::phase0::operations::PendingAttestation;
 use crate::phase0::state_transition::{
-    get_block_root, get_current_epoch, get_previous_epoch, Context, Error,
+    get_block_root, get_current_epoch, get_previous_epoch, get_randao_mix, Context, Error,
 };
 use crate::primitives::Epoch;
 use ssz_rs::prelude::*;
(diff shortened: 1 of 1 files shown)


ralexstokes:
> Also lmk if I should keep the PR's specific to a fn, or if I should group them together in a PR so that I'm not spamming with PRs consisting of a few lines of code.

matter of style, and personally i prefer one PR per "topic"... so even if these are on the small side i feel it accelerates the review process and keeps a good cadence on overall progress with #29.  some of the "subtransitions" will land much bigger PRs -- someone, perhaps even you, will get to those and there is something to be said for getting a bunch of "small wins" w/ several merged PRs to tee you up for something like implementing the rewards processing

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -282,8 +282,13 @@ pub fn process_randao_mixes_reset<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<(), Error> {
+    let current_epoch = get_current_epoch(state, context);
+    let next_epoch = get_current_epoch(state, context) + 1;
```
ralexstokes, line 288: can save the overhead of a fn call here
```suggestion
    let next_epoch = current_epoch + 1;
```

ralexstokes (commented):
nice job! i left one small suggestion

the `clone` on the `&Bytes32` is good -- we actually want a distinct copy as the new `Bytes32` instance will accumulate mutations with each block throughout the next epoch and if we shared them somehow, it would lead to a bad state.

PR #34, opened 2021-12-26
add fn body for `process_block_header`

Cravtos:


src/phase0/state_transition/block_processing.rs
```diff
@@ -160,16 +163,38 @@ fn process_block_header<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    _block: &mut BeaconBlock<
+    block: &mut BeaconBlock<
```
ralexstokes, line 166: this `&mut` was an accident in https://github.com/ralexstokes/ethereum_consensus/commit/68a652bbfc269ff142c7b36e97c78b6d7ae68eaa and should just be `&`. 

i just fixed it on `main` via #35 which you can pull into this branch if you rebase

src/phase0/state_transition/mod.rs
```diff
@@ -60,6 +60,18 @@ pub enum Error {
         previous: Epoch,
         current: Epoch,
     },
+    #[error("invalid block header: {0}")]
+    InvalidHeader(InvalidHeader),
```
ralexstokes, line 64: this is likely not very clear but it would have been conceptually cleaner in the `phase0` spec if we had put the things like the block header processing, the randao processing and the eth1 data processing into the normal "operation processing" flow along w/ the rest of the block.

i have been leaning towards this route in this repo which would suggest moving `InvalidHeader` into a variant of `InvalidOperation`.  this will add one more enum constructor into the library code but i'd prefer the symmetry it yields.

src/phase0/state_transition/mod.rs
```diff
@@ -60,6 +60,18 @@ pub enum Error {
...
+}
+
+#[derive(Debug, Error)]
+pub enum InvalidHeader {
```
ralexstokes, line 68: this does increase the verbosity but we will be happier later when there are other types of `Header`s floating around (e.g. a `ShardHeader`).
```suggestion
pub enum InvalidBeaconBlockHeader {
```

src/phase0/state_transition/mod.rs
```diff
@@ -60,6 +60,18 @@ pub enum Error {
...
+    #[error("slots mismatch")]
+    StateSlotMismatch,
```
ralexstokes, line 70: i'd suggest we put as much information into the error as we have from the context in which it is created:


```suggestion
    #[error("mismatch between state slot {state_slot} and block slot {block_slot}")]
    StateSlotMismatch { state_slot: Slot, block_slot: Slot },
```

src/phase0/state_transition/mod.rs
```diff
@@ -60,6 +60,18 @@ pub enum Error {
...
+    #[error("roots mismatch")]
+    ParentBlockRootMismatch,
```
ralexstokes, line 72:
```suggestion
    #[error("mismatch between the block's parent root {provided} and the expected parent root {expected}")]
    ParentBlockRootMismatch { expected: Root, provided: Root },
```

src/phase0/state_transition/mod.rs
```diff
@@ -60,6 +60,18 @@ pub enum Error {
...
+    #[error("proposer is slashed")]
+    ProposerSlashed,
```
ralexstokes, line 74:
```suggestion
    #[error("proposer with index {0} is slashed")]
    ProposerSlashed(ValidatorIndex),
```

src/phase0/state_transition/block_processing.rs
```diff
@@ -160,16 +163,38 @@ fn process_block_header<
...
+    if block.slot != state.slot {
+        return Err(Error::InvalidHeader(InvalidHeader::StateSlotMismatch));
+    }
+
+    let expected_block_root = state.latest_block_header.hash_tree_root()?;
+    if block.parent_root != expected_block_root {
+        return Err(Error::InvalidHeader(InvalidHeader::ParentBlockRootMismatch));
+    }
+
+    state.latest_block_header = crate::phase0::beacon_block::BeaconBlockHeader {
```
ralexstokes, line 185:
```suggestion
    state.latest_block_header = BeaconBlockHeader {
```

and up above let's just import it as a single symbol:

`use crate::phase0::beacon_block::BeaconBlockHeader`

src/phase0/state_transition/block_processing.rs
```diff
@@ -160,16 +163,38 @@ fn process_block_header<
...
         MAX_DEPOSITS,
         MAX_VOLUNTARY_EXITS,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<(), Error> {
+    if block.slot != state.slot {
+        return Err(Error::InvalidHeader(InvalidHeader::StateSlotMismatch));
+    }
+
```
ralexstokes, line 179: missing some validations here:

- block slot is newer than the latest header in the state
- block's proposer is equal to the one expected based on our local state

src/phase0/state_transition/block_processing.rs
```diff
@@ -160,16 +163,38 @@ fn process_block_header<
...
+    let expected_block_root = state.latest_block_header.hash_tree_root()?;
+    if block.parent_root != expected_block_root {
```
ralexstokes, line 181: imo can be a bit more precise here:

```suggestion
    let expected_parent_root = state.latest_block_header.hash_tree_root()?;
    if block.parent_root != expected_parent_root {
```

src/phase0/state_transition/block_processing.rs
```diff
@@ -160,16 +163,38 @@ fn process_block_header<
...
+        return Err(Error::InvalidHeader(InvalidHeader::ParentBlockRootMismatch));
+    }
+
+    state.latest_block_header = crate::phase0::beacon_block::BeaconBlockHeader {
+        slot: block.slot,
```
ralexstokes, line 186: missing proposer index

src/phase0/state_transition/block_processing.rs
```diff
@@ -160,16 +163,38 @@ fn process_block_header<
...
+        parent_root: block.parent_root,
+        body_root: block.body.hash_tree_root()?,
+        ..Default::default()
+    };
+
+    let proposer_index = get_beacon_proposer_index(state, context)?;
```
ralexstokes, line 192: can hoist this up above in the relevant validations

ralexstokes (changes requested):
welcome @Cravtos and excellent job on  this PR!

I like how you were careful to make a precise error variant for each failure mode of the validations.

I left some feedback below on a variety of things; the main blocker being that some of the logic is missing from the implementation of `process_block_header`. Also, you'll want to rebase against the change made in #35.

src/phase0/state_transition/block_processing.rs
Cravtos: My bad, I was looking at the old version of the specification. Added here: 7b8dfbf

src/phase0/state_transition/block_processing.rs
Cravtos: Done 7b8dfbf

src/phase0/state_transition/mod.rs
Cravtos: Moved 7b8dfbf

src/phase0/state_transition/block_processing.rs
Cravtos: Fixed 7b8dfbf

src/phase0/state_transition/block_processing.rs
Cravtos: Done. 
But now there is an issue with `block.body.hash_tree_root()` since `hash_tree_root()` takes `self` as `&mut`. I need clues on how can I work around this issue.
I don't understand, why `hash_tree_root()` wants mutable reference?

src/phase0/state_transition/mod.rs
```diff
@@ -82,6 +82,28 @@ pub enum InvalidOperation {
...
+    #[error("invalid block header: {0}")]
+    HeaderValidation(InvalidBeaconBlockHeader),
+}
+
+#[derive(Debug, Error)]
+pub enum InvalidBeaconBlockHeader {
+    #[error("mismatch between state slot {state_slot} and block slot {block_slot}")]
+    StateSlotMismatch { state_slot: Slot, block_slot: Slot },
+    #[error("mismatch between the block's parent root {expected} and the expected parent root {provided}")]
+    ParentBlockRootMismatch { expected: Root, provided: Root },
```
Cravtos, line 94: It doesn't compile since `ssz_rs::Node` does not satisfy `std::fmt::Display`. Should I implement it or it is better not to touch the `ssz_rs` code?

src/phase0/state_transition/mod.rs
Cravtos: What about this enum structure, as in https://github.com/ralexstokes/ethereum_consensus/pull/40?
```rust
#[derive(Debug, Error)]
pub enum InvalidOperation {
    // --snip--
    #[error("invalid block: {0}")]
    BlockProcessing(InvalidBlock),
}

#[derive(Debug, Error)]
pub enum InvalidBlock {
    #[error("invalid randao signature")]
    RandaoSignatureInvalid,
    #[error("mismatch between state slot {state_slot} and block slot {block_slot}")]
    StateSlotMismatch { state_slot: Slot, block_slot: Slot },
    #[error("mismatch between the block's parent root {expected} and the expected parent root {provided}")]
    ParentBlockRootMismatch { expected: Root, provided: Root },
    // and so on...
}
```

src/phase0/state_transition/block_processing.rs
ralexstokes: ah! sorry i mislead you here....

i did think the mutability on `block` was an accident (bc i had just blindly copied it the first time) but we do actually need it for the reason you point you -- taking the `hash_tree_root`.

and the reason why we have `hash_tree_root(&mut self)` is that SSZ types have internal caching that may need to update internal state when getting the hash tree root (edits to some SSZ type do not immediately update the backing Merkle tree to avoid paying the hashing cost if you don't need it)

i have a WIP PR of caching for `Vector`s here if you want to see more: https://github.com/ralexstokes/ssz_rs/pull/6

src/phase0/state_transition/mod.rs
ralexstokes: see my other comment  in #40: https://github.com/ralexstokes/ethereum_consensus/pull/40#discussion_r776394266

i think it is nicer to just keep the symmetry of each of the block sub-transitions as their own enum variants

src/phase0/state_transition/mod.rs
```diff
@@ -70,6 +70,28 @@ pub enum InvalidOperation {
     IndexedAttestation(InvalidIndexedAttestation),
     #[error("invalid deposit: {0}")]
     Deposit(InvalidDeposit),
+    #[error("invalid block header: {0}")]
+    HeaderValidation(InvalidBeaconBlockHeader),
```
ralexstokes, line 74: i think we can simplify the name here:
```suggestion
    Header(InvalidBeaconBlockHeader),
```

src/phase0/state_transition/mod.rs
ralexstokes: if you wanted more structure here i would actually invert them:

`phase0::state_transition::Error`

has a new variant:

`InvalidBlock(InvalidBlock)`

that looks like:

```rust
pub enum InvalidBlock {
    Header(InvalidBeaconBlockHeader),
    InvalidOperation(InvalidOperation),
```

where we moved the `InvalidOperation` from the top-level down into the `InvalidBlock` nesting

what do you think about that? i actually like it better once i type it out!

ralexstokes: i'd also introduce some helper functions to construct the error variants if we go this route to avoid heavily nested enum literals...

you could either just make functions like `block_error(...)` or `operation_error(...)` or try another route to change what error is collected in `process_block` and massage upon returning so that you end up with a `state_transition::Error`

src/phase0/state_transition/block_processing.rs
```diff
@@ -160,16 +165,69 @@ fn process_block_header<
...
+    if block.slot <= state.latest_block_header.slot {
+        return Err(Error::InvalidOperation(InvalidOperation::HeaderValidation(
+            InvalidBeaconBlockHeader::OlderThanLatestBlockHeader {
+                block_slot: block.slot,
+                latest_block_header_slot: state.latest_block_header.slot,
+            },
+        )));
+    }
+
+    let state_proposer_index = get_beacon_proposer_index(state, context)?;
```
ralexstokes, line 196: nitpick but
```suggestion
    let proposer_index = get_beacon_proposer_index(state, context)?;
```

src/phase0/state_transition/mod.rs
ralexstokes: ah, let's just use `Debug` here

```suggestion
    #[error("mismatch between the block's parent root {expected:?} and the expected parent root {provided:?}")]
    ParentBlockRootMismatch { expected: Root, provided: Root },
```

ralexstokes (commented):
sweet updates!

left some more comments

src/phase0/state_transition/block_processing.rs
Cravtos: Got it, thank you! Used `&mut`: b7cd5e2

src/phase0/state_transition/block_processing.rs
Cravtos: Renamed here: 89fffa6

src/phase0/state_transition/mod.rs
Cravtos: Simplified: 89fffa6

src/phase0/state_transition/mod.rs
Cravtos: Done: 5b790a5

src/phase0/state_transition/mod.rs
Cravtos: I like it too. Could do that in a separate PR

Cravtos: @ralexstokes, look at it here: https://github.com/Cravtos/ethereum_consensus/pull/2

src/phase0/state_transition/block_processing.rs
```diff
@@ -160,16 +165,69 @@ fn process_block_header<
...
+            },
+        )));
+    }
+
+    let proposer_index = get_beacon_proposer_index(state, context)?;
+    if block.proposer_index != proposer_index {
+        return Err(Error::InvalidOperation(InvalidOperation::Header(
+            InvalidBeaconBlockHeader::ProposerIndexMismatch {
+                block_proposer_index: block.proposer_index,
+                state_proposer_index: proposer_index,
```
ralexstokes, line 201: i think in this context `state` is redundant so in favor of reducing verbosity:

```suggestion
                proposer_index,
```

src/phase0/state_transition/mod.rs
```diff
@@ -70,6 +70,28 @@ pub enum InvalidOperation {
...
+    ProposerSlashed(ValidatorIndex),
+    #[error("block slot {block_slot} is older than the latest block header slot {latest_block_header_slot}")]
+    OlderThanLatestBlockHeader {
+        block_slot: Slot,
+        latest_block_header_slot: Slot,
+    },
+    #[error("mismatch between the block proposer index {block_proposer_index} and the state proposer index {state_proposer_index}")]
+    ProposerIndexMismatch {
+        block_proposer_index: usize,
+        state_proposer_index: usize,
```
ralexstokes, line 93: i think we can do this (for a bit more type safety):


```suggestion
        block_proposer_index: ValidatorIndex,
        state_proposer_index: ValidatorIndex,
```

ralexstokes (commented):
nice work!

i left two minor suggestions -- this should be ready to merge soon!

src/phase0/state_transition/mod.rs
Cravtos: Done: 24b9f6e

src/phase0/state_transition/block_processing.rs
Cravtos: Done:  24b9f6e

ralexstokes (approved):
nice work!

PR #36, opened 2021-12-27
feat: add fn body for 'process_participation_record_updates'

buchhlz2:
@ralexstokes before I merge this -- since you were just creating skeletons, should I remove `_context` from the fn parameters if it's unused? Or, are you trying to keep everything sorta consistent with `state` and `context` always being passed?

### DIFF
--- a/src/phase0/state_transition/epoch_processing.rs
+++ b/src/phase0/state_transition/epoch_processing.rs
@@ -288,7 +288,7 @@ pub fn process_randao_mixes_reset<
     context: &Context,
 ) -> Result<(), Error> {
     let current_epoch = get_current_epoch(state, context);
-    let next_epoch = get_current_epoch(state, context) + 1;
+    let next_epoch = current_epoch + 1;
 
     state.randao_mixes[next_epoch as usize % context.epochs_per_historical_vector] =
         get_randao_mix(state, current_epoch).clone();
--- a/src/phase0/state_transition/epoch_processing.rs
+++ b/src/phase0/state_transition/epoch_processing.rs
@@ -5,6 +5,7 @@ use crate::phase0::state_transition::{
 };
 use crate::primitives::Epoch;
 use ssz_rs::prelude::*;
+use std::mem;
 
 pub fn get_matching_source_attestations<
     'a,
@@ -340,11 +341,10 @@ pub fn process_participation_record_updates<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    _context: &Context,
 ) -> Result<(), Error> {
-    state.previous_epoch_attestations = state.current_epoch_attestations.clone();
-    state.current_epoch_attestations.clear();
-    
+    let current_attestations = mem::take(&mut state.current_epoch_attestations);
+    state.previous_epoch_attestations = current_attestations;
+
     Ok(())
 }
 
@@ -379,6 +379,6 @@ pub fn process_epoch<
     process_slashings_reset(state, context)?;
     process_randao_mixes_reset(state, context)?;
     process_historical_roots_update(state, context)?;
-    process_participation_record_updates(state, context)?;
+    process_participation_record_updates(state)?;
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


ralexstokes:
> should I remove _context from the fn parameters if it's unused? 

yes please!

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -342,6 +342,9 @@ pub fn process_participation_record_updates<
     >,
     _context: &Context,
 ) -> Result<(), Error> {
+    state.previous_epoch_attestations = state.current_epoch_attestations.clone();
+    state.current_epoch_attestations.clear();
```
ralexstokes, line 346: so this is a _correct_ implementation but i will suggest an optimization because it is pretty straightforward to add here:

what you have now:
1) clone the `current` data (so now there are *two* copies of the same data sitting on the heap)
2) move the new copy into the part of the state's memory for the `previous` data (and clear the previous `previous` data)
3) clear the first copy of the `current` data

what i would like to improve upon is making a second copy of the `current` attestations before writing them into the `previous` field; we don't need two copies, we just need to *move* the single copy into another place.

and if you try to write something along these lines, then you will get a compiler error as what we are trying to do is actually forbidden by "safe" rust. so then we have to use "unsafe" rust to do what we want. and luckily this type of operation is common enough that there is a nice (safe) function for it in the standard library under `std::mem::take`

https://doc.rust-lang.org/stable/std/mem/fn.take.html

if you look at the second example, you will see it is exactly our problem that they call out!

so i'd try something like this:

```suggestion
    // you will need to import this (up above)
    use std::mem;
    
    let current_attestations = mem::take(&mut state.current_epoch_attestations);
    state.previous_epoch_attestations = current_attesations;
```

i believe we could go even further and actually leave the underlying `Vec`s in place and just move the backing buffer but that would require some further unsafe spelunking and starts to over-optimize at the cost of readability/understandability imo -- if we ever profile this code and find a hot-spot here we can discuss deeper options. i think this optimization is worth it as `mem::take` semantics are pretty clear imo and we don't have to actually write an `unsafe` block as `mem::take` provides a nice "safe" porcelain for us... and given the usage pattern of this field (every epoch, add a bunch of data to this field and then on every epoch boundary rotate the fields) it is worth implementing a basic optimization with `mem::take` to avoid a duplicate copy.

ralexstokes (commented):
excellent work!

two comments to consider before merging this in!

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Interesting -- agreed, and using `take()` definitely makes sense here; wasn't aware of its functionality! And after reading the docs, there's also a similar `swap()` function...any reason not to use that instead since it's semantically similar? Something like:
```
mem::swap(
        &mut state.previous_epoch_attestations,
        &mut state.current_epoch_attestations,
);
state.current_epoch_attestations.clear();
```
`swap()` is unstable with `const`s but that shouldn't be a problem here.

ralexstokes: that is another possible implementation although in my opinion `take` is more semantically correct than `swap` as it essentially means what you have written: "move `a` into `b` and make `a` some 'default' impl"

some additional color:

by putting it behind one API (the function call to `mem::take`) there is now one place where this operation can be optimized -- and i don't know but could expect situations where the compiler is able to produce more optimal code given the invariants of `take` that it would not be able to deduce from the `swap + clear` approach.

by leveraging the `Default` impl we could now conceivably produce a hyper-optimized `Default` impl, perhaps one that has a custom allocator or $INSERT_FEATURE to increase performance -- and it may not be exactly for this field in this struct or for this type... but generically speaking, `Default` is a better semantic target for such a thing than a bespoke method like `clear`

and we can argue about this one, but on an aesthetic  front, `take + move` feels better to me than `swap + clear` given the "affine" nature of Rust in general

buchhlz2: Ohhhh, I get it now -- `take()` takes (duh) that value _and_ handles the desired clearing via `Default`, in one swoop. Using `swap()` is basically what I originally attempted w/o the cloning aspect. Cool.

PR #37, opened 2021-12-27
feat: add fn 'process_historical_roots_update'

buchhlz2:


src/phase0/state_transition/epoch_processing.rs
```diff
@@ -315,8 +315,18 @@ pub fn process_historical_roots_update<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<(), Error> {
+    let next_epoch = get_current_epoch(state, context) + 1;
+    if (next_epoch % (SLOTS_PER_HISTORICAL_ROOT as u64 / context.slots_per_epoch)) == 0 {
```
buchhlz2, line 321: Specs call out the division should be a floor (`//`) -- I originally thought Rust's `u64` does this automatically, but I know Lighthouse uses some custom macros/methods called `safe_div`. And now I'm seeing some nightly/unstable methods do floor division, so I might need to refactor.

Is converting to an `f64` to use its `floor` method, and then back to a `u64` too messy?

buchhlz2: Okay -- so nvm, I wrote a quick test to validate (should have done this prior), and `u64` does properly round to floor. Idk if something like `safe_div` is needed, so I'll assume simply using `/` should suffice.

ralexstokes: right, the arithmetic operators on unsigned integer types in rust do the integer operations.

there are some corner cases to watch out for and this is why lighthouse uses the `safe_*` methods.

i am certainly open to making a "safety" pass on the code to use these sorts of methods but lean towards this repo having as little additional conceptual overhead as possible to facilitate understandability / readability.

and to that end i have been trying to stick to the plain arithmetic operators where possible.

if you feel strongly, i'd open an issue and we can move the conversation there

ralexstokes, line 321: this is not very clear and i even started out using the const values like you did to cut down on needing to pass around a `Context` and hopefully move some computation from runtime to compile-time.

however, i think at the moment for consistency/uniformity i'm keeping the const params at the type-level and where needed referring to values from the `context`

so then this would look like:

```suggestion
    let epochs_per_historical_root = context.slots_per_historical_root / context.slots_per_epoch;
    if next_epoch % epochs_per_historical_root == 0 {
```

where i pulled out the boundary computation for readability (and dropped what i think are extra parens) and also implied a change in the `Context` where the `slots_per_historical_root` field has type `Slot` (it has type `usize` now)

so also this change:

```rust
    pub slots_per_historical_root: Slot,
```

i don't think this type change will break anything as the other constraint is needed `usize` at the type-level but that is handled by the const generic param.

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -315,8 +315,18 @@ pub fn process_historical_roots_update<
...
+        let mut historical_batch = HistoricalBatch {
+            block_roots: state.block_roots.clone(),
+            state_roots: state.state_roots.clone(),
```
ralexstokes, line 324: similarly to this: https://github.com/ralexstokes/ethereum_consensus/pull/36#discussion_r775979388

i'm going to want to avoid these `clone`s :)

if we do the most straightforward thing to remove the calls to `clone`, these fields then become references:

```rust
#[derive(Default, Debug, SimpleSerialize)]
pub struct HistoricalBatch<const SLOTS_PER_HISTORICAL_ROOT: usize> {
    pub block_roots: &Vector<Root, SLOTS_PER_HISTORICAL_ROOT>,
    pub state_roots: &Vector<Root, SLOTS_PER_HISTORICAL_ROOT>,
}
```

and we hit a problem as we can no longer derive `Default` or `SimpleSerialize`: there is no `Default` for `&T` as that would likely just point to "unsafe" uninitialized memory, and `SimpleSerialize` needs `Default` right now for the {de,}serialization routines.

so rather than make broad changes to `ssz_rs` at the moment (e.g. we could try to relax the `Default` constraint, which will be tricky, or we could pull out the merkleization functionality into a separate trait (but then need to implement `derive` functionality, etc.)), we can just leverage what we know about SSZ to work around this problem at the cost of a small deviation from the spec.

the `HistoricalBatch` container is essentially just a wrapper around the `block_roots` and `state_roots` data to get a nice "summary" merkle root over the entire set of block and state roots. and given that we just need this one root, we can basically perform part of the Merkleization manually and then use `ssz_rs` to do the rest. note: if we needed to serialize the `HistoricalBatch` data we could not take advantage of this but luckily we just need the root!

to finish the thought, we make some changes like this:

```rust
// 1. change definition
// and perhaps leave a helpful comment (or at least link this github comment!) to document why we deviate from the spec here
#[derive(Default, Debug, SimpleSerialize)]
pub struct HistoricalBatchAccumulator {
    pub block_roots_root: Root,
    pub state_roots_root: Root,
}

// 2. and then in this function:
let mut historical_batch = HistoricalBatchAccumulator {
    block_roots_root: state.block_roots.hash_tree_root()?,
    state_roots_root: state.state_roots.hash_tree_root()?,
};
```

src/phase0/state_transition/epoch_processing.rs
buchhlz2: All for simplicity...can do that if there's a reason one day but yeah definitely no need for it.

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Got it -- makes sense. I've made the necessary changes and want to call out one minor casting needed in `Context` after the `slots_per_historical_root` type change...which I think is the desired approach but double checking: https://github.com/buchhlz2/ethereum_consensus/commit/6d0b56b1807ce82606dd1dbd9de2840a86112f99#r62609271

src/phase0/state_transition/context.rs
```diff
@@ -69,7 +69,7 @@ impl Context {
             max_seed_lookahead: preset.max_seed_lookahead,
             min_epochs_to_inactivity_penalty: preset.min_epochs_to_inactivity_penalty,
             epochs_per_eth1_voting_period: preset.epochs_per_eth1_voting_period,
-            slots_per_historical_root: preset.slots_per_historical_root,
+            slots_per_historical_root: preset.slots_per_historical_root as Slot,
```
buchhlz2, line 72: If I don't cast this as a `Slot`, then I'd have to mess with the `const`s in `presets/mainnet.rs` (e.g., change `SLOTS_PER_HISTORICAL_ROOT` from `usize` to `u64`). Just calling out if there's a reason this isn't ideal, but the former was obviously more straightforward than the latter.

src/phase0/presets/mainnet.rs
```diff
@@ -99,7 +99,7 @@ pub const PRESET: Preset = Preset {
 
 pub type IndexedAttestation = operations::IndexedAttestation<MAX_VALIDATORS_PER_COMMITTEE>;
 pub type PendingAttestation = operations::PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>;
-pub type HistoricalBatch = beacon_state::HistoricalBatch<SLOTS_PER_HISTORICAL_ROOT>;
+pub type HistoricalBatchAccumulator = beacon_state::HistoricalBatchAccumulator;
```
ralexstokes, line 102: we don't need this type defn anymore!

the idea w/ the types in the various `presets` is just to "specialize" them so users of this library don't see the generic parameters.

if we move ahead w/ this new accumulator type then we can just delete this line!

src/phase0/state_transition/context.rs
ralexstokes: yep nice catch....

so ideally we would not need to be so precise w/ the type of the const generic and that is what i was getting at here: https://github.com/ralexstokes/ethereum_consensus/issues/1

but until we can unify the numeric types somehow (if ever), we will need a cast....

so -- right idea but i would suggest we move this "up" one struct to the `Preset`

so there is the constant `pub const ...: usize = ...;` and then the field in the `Preset` is also a `Slot` and you will need to do this `as Slot` cast there.

i think there are a few places where we can tighten the typing in `Context` / `Preset` (bc they keep cropping up as we implement the rest of the spec) so if you'd like to just raise an issue and tag me i can make a pass across the entire struct -- it seems like something i should have done the first time

ralexstokes (commented):
nice updates!

Issue #38, opened 2021-12-29
Add stronger typing to `Context` & `Preset`


buchhlz2:
While implementing the various functions within the `state_transition` crate, a number of fields have required stronger typing than what currently exists. For example, https://github.com/ralexstokes/ethereum_consensus/pull/31 required `epochs_per_eth1_voting_period` to be updated from `usize` to `Epoch`.

Thus, the remaining types can also benefit from stronger typing within both of these structs as this will likely continue to occur with further code updates to the `state_transition` and other crates.

PR #40, opened 2021-12-29
add fn body for `process_randao`

Cravtos:


src/phase0/state_transition/mod.rs
```diff
@@ -70,6 +70,14 @@ pub enum InvalidOperation {
     IndexedAttestation(InvalidIndexedAttestation),
     #[error("invalid deposit: {0}")]
     Deposit(InvalidDeposit),
+    #[error("invalid block: {0}")]
+    BlockProcessing(InvalidBlock),
```
ralexstokes, line 74: while it may not exactly be called an "operation" in the spec, i think just having a variant for randao failures is preferred to keep the symmetry w/ the other errors here

and we can inline the one variant of randao processing w/ the invalid signature


```suggestion
    #[error("invalid randao (BLS signature): {0}")]
    Randao(BLSSignature),
```
i'll leave it up to you if you want to include more detailed error information in the enum variant, it could contain pubkey, message and signature but that could be overkill -- you pick

src/phase0/state_transition/block_processing.rs
```diff
@@ -198,16 +202,36 @@ fn process_randao<
...
+
+    let proposer_index = get_beacon_proposer_index(state, context)?;
+    let proposer = &state.validators[proposer_index];
+
+    let domain = get_domain(state, DomainType::Randao, Some(epoch), context)?;
+    let signing_root = compute_signing_root(&mut epoch, domain)?;
+
+    if !body
+        .randao_reveal
+        .verify(&proposer.pubkey, signing_root.as_ref())
```
ralexstokes, line 225: nitpick, but i think `as_bytes` is a bit more semantically aligned w/ what `verify` wants, rather than simply a reference to the root

ralexstokes (commented):
excellent work!

i left some minor comments and think we can simplify the error handling

src/phase0/state_transition/block_processing.rs
```diff
@@ -198,16 +202,36 @@ fn process_randao<
...
+    {
+        return Err(Error::InvalidOperation(InvalidOperation::BlockProcessing(
+            InvalidBlock::RandaoSignatureInvalid,
+        )));
+    }
+
+    let mix = get_randao_mix(state, epoch).xor(hash(body.randao_reveal.as_bytes()));
+    state.randao_mixes[epoch as usize % EPOCHS_PER_HISTORICAL_VECTOR] = mix;
```
ralexstokes, line 233: see the first part of this comment: https://github.com/ralexstokes/ethereum_consensus/pull/37#discussion_r775984880

i aim to maintain consistency across uses like these :)

```suggestion
    state.randao_mixes[epoch % context.epochs_per_historical_vector] = mix;
```

ralexstokes (commented):
github dropped my other review comment!

src/phase0/state_transition/block_processing.rs
Cravtos: Fixed: dc8ad40

src/phase0/state_transition/block_processing.rs
Cravtos: Fixed: dc8ad40

src/phase0/state_transition/mod.rs
Cravtos: Hope enum naming is alright, let me know if not: 62eaaca

src/phase0/state_transition/block_processing.rs
```diff
@@ -198,16 +202,35 @@ fn process_randao<
...
+        .randao_reveal
+        .verify(&proposer.pubkey, signing_root.as_bytes())
+    {
+        return Err(Error::InvalidOperation(InvalidOperation::Randao(
+            BLSSignature::InvalidSignature,
+        )));
+    }
+
+    let mix = get_randao_mix(state, epoch).xor(hash(body.randao_reveal.as_bytes()));
+    state.randao_mixes[epoch as usize % context.epochs_per_historical_vector] = mix;
```
ralexstokes, line 233: making a note to myself patch up the cast as part of #38.

src/phase0/state_transition/mod.rs
ralexstokes: looks good!

although i meant that we can just use the actual signature data, not the most inner enum provided in 62eaaca

to do this we would delete the enum `BLSSignature` and adjust the `InvalidOperation::Randao` variant to just wrap a `crypto.Signature` (from the top-level module `crypto`)

```rust
// modify the import at the top:
use crate::crypto::{fast_aggregate_verify, hash, Signature as BLSSignature};

pub enum InvalidOperation {
    // other variants elided
    Randao(BLSSignature),
}
```

where i added the signature alias `BLSSignature` just to disambiguate from some other type of generic `Signature` that could be floating around

ralexstokes: and then at the call site something like:

```rust
        return Err(Error::InvalidOperation(InvalidOperation::Randao(
            body.randao_reveal.clone(),
        )));
```

then at the least a user of this library will see the faulty signature/reveal upon error

ralexstokes (commented):
left a comment, looking good!

src/phase0/state_transition/mod.rs
Cravtos: Fixed: 1abdf5c

ralexstokes (approved):
excellent updates!

PR #41, opened 2022-01-05
Add fn body for `process_eth1_data`

Cravtos:


src/phase0/state_transition/block_processing.rs
```diff
@@ -236,16 +236,28 @@ fn process_eth1_data<
...
         MAX_PROPOSER_SLASHINGS,
         MAX_VALIDATORS_PER_COMMITTEE,
         MAX_ATTESTER_SLASHINGS,
         MAX_ATTESTATIONS,
         MAX_DEPOSITS,
         MAX_VOLUNTARY_EXITS,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<(), Error> {
```
Cravtos, line 248: Should I remove `Result<(), Error>` from the function return type if no error can be returned?

ralexstokes: yes please!

src/phase0/state_transition/block_processing.rs
```diff
@@ -236,16 +236,28 @@ fn process_eth1_data<
...
+    state.eth1_data_votes.push(body.eth1_data.clone());
+
+    let votes_count = state
+        .eth1_data_votes
+        .iter()
+        .filter(|&vote| *vote == body.eth1_data)
```
ralexstokes, line 254: nitpick: 

i believe we can just do this
```suggestion
        .filter(|vote| *vote == body.eth1_data)
```

and there are sufficient blanket impls of `PartialEq` for refs that the compiler will do the right thing

and then we have less code to grok

ralexstokes (commented):
looks great! i left one minor suggestion

we can imagine a more efficient implementation than this but for the time being let's just stick to the spec :)

src/phase0/state_transition/block_processing.rs
Cravtos: The issue is 'body.eth1_data' is not a ref. 

I could do
```suggestion
        .filter(|vote| **vote == body.eth1_data)
```
or
```suggestion
        .filter(|vote| *vote == &body.eth1_data)
```
but would that change anything?

src/phase0/state_transition/block_processing.rs
Cravtos: Done: da0a232

src/phase0/state_transition/block_processing.rs
ralexstokes: ah good point, i was confused about the `body` item!

ralexstokes (approved):
nice updates!

PR #42, opened 2022-01-05
feat: add fn body to `process_registry_updates`

buchhlz2:
This PR is a WIP as I'm having some issues with Rust and am looking for feedback.

I'm trying to implement [`process_registry_updates`](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#registry-updates), which has 3 primary steps:

- [x] Process activation eligibility and ejections
- [x] Queue validators eligible for activation and not yet dequeued for activation
- [x] Dequeued validators for activation up to churn limit

I think I have the first step completed (obviously open to feedback), but I've had issues on the next step while trying to recreate python's lambda & sorting. This is what I've tried to do:
1. Get validators from state
2. Create an iterator and enumerate over the values to get a tuple of `(ValidatorIndex, &Validator)`
3. Sort the values in place (I think this is what `sort_by` does) with priority of `activation_eligibility_epoch`, then validator's `index` (where the variables in the closure could be removed but are created for clarity)
4. Map the values so that the final result is just a vector of `ValidatorIndex`s

I'm running into issues somewhere prior to step 4 because in the code's current state (stops after step 3), the type on `activation_queue` is simply `()`...instead of a vector of `(ValidatorIndex, &Validator)`. I'm also not sure if using `sort_by` is the only option here but wanted to avoid using other crates that accomplish this.

### DIFF
--- a/src/phase0/state_transition/epoch_processing.rs
+++ b/src/phase0/state_transition/epoch_processing.rs
@@ -187,7 +187,6 @@ pub fn process_registry_updates<
             }
         })
         .collect::<Vec<ValidatorIndex>>();
-        
     // Order by the sequence of activation_eligibility_epoch setting and then index
     activation_queue.sort_by(|&i, &j| {
         let a = &state.validators[i];
--- a/src/phase0/state_transition/epoch_processing.rs
+++ b/src/phase0/state_transition/epoch_processing.rs
@@ -1,14 +1,14 @@
 use crate::phase0::beacon_state::{BeaconState, HistoricalBatchAccumulator};
 use crate::phase0::operations::PendingAttestation;
 use crate::phase0::state_transition::{
-    decrease_balance, get_block_root, get_current_epoch, get_previous_epoch, get_randao_mix,
-    get_total_active_balance, initiate_validator_exit, is_active_validator,
-    is_eligible_for_activation, is_eligible_for_activation_queue, Context, Error,
+    compute_activation_exit_epoch, decrease_balance, get_block_root, get_current_epoch,
+    get_previous_epoch, get_randao_mix, get_total_active_balance, get_validator_churn_limit,
+    initiate_validator_exit, is_active_validator, is_eligible_for_activation,
+    is_eligible_for_activation_queue, Context, Error,
 };
-use crate::phase0::validator::Validator;
 use crate::primitives::{Epoch, Gwei, ValidatorIndex};
 use ssz_rs::prelude::*;
-use std::{cmp, mem};
+use std::mem;
 
 pub fn get_matching_source_attestations<
     'a,
@@ -175,43 +175,28 @@ pub fn process_registry_updates<
     }
 
     // Queue validators eligible for activation and not yet dequeued for activation
-    let _activation_queue = state
+    let mut activation_queue = state
         .validators
         .iter()
         .enumerate()
         .filter(|&(_, validator)| is_eligible_for_activation(state, validator))
-        .collect::<Vec<(ValidatorIndex, &Validator)>>()
-        // Order by the sequence of activation_eligibility_epoch setting and then index
-        .sort_by(|a, b| {
-            let a_activation_eligibility_epoch = a.1.activation_eligibility_epoch;
-            let a_index = a.0;
-            let b_activation_eligibility_epoch = b.1.activation_eligibility_epoch;
-            let b_index = b.0;
-            match a_activation_eligibility_epoch.cmp(&b_activation_eligibility_epoch) {
-                cmp::Ordering::Less => cmp::Ordering::Less,
-                cmp::Ordering::Greater => cmp::Ordering::Greater,
-                cmp::Ordering::Equal => match a_index.cmp(&b_index) {
-                    cmp::Ordering::Less => cmp::Ordering::Less,
-                    cmp::Ordering::Greater => cmp::Ordering::Greater,
-                    // TODO: throw some error if the valators' indices are equal -- should not be possible
-                    cmp::Ordering::Equal => cmp::Ordering::Equal,
-                },
-            }
-        });
-    // TODO: map sorted vec from (ValidatorIndex, &Validator) to just the ValidatorIndex
-    // .map(|(index, _)| index);
-    // Having issue where `activation_queue` is defined as `()` instead of a vector, so no operations can be performed
+        .map(|(i, _)| i)
+        .collect::<Vec<ValidatorIndex>>();
+    // Order by the sequence of activation_eligibility_epoch setting and then index
+    activation_queue.sort_by(|i, j| {
+        let a = &state.validators[*i];
+        let b = &state.validators[*j];
+        (a.activation_eligibility_epoch, i).cmp(&(b.activation_eligibility_epoch, j))
+    });
 
-    // TODO: WIP on below code; must finish `activation_queue` issues first
-    // Reminder: import `compute_activation_exit_epoch`, `get_validator_churn_limit` from `state_transition`
     // Dequeued validators for activation up to churn limit
-    // for i in activation_queue
-    //     .iter()
-    //     .take(get_validator_churn_limit(state, context) as usize)
-    // {
-    //     let validator = &mut state.validators[i.clone()];
-    //     validator.activation_epoch = compute_activation_exit_epoch(current_epoch, context);
-    // }
+    for i in activation_queue
+        .iter()
+        .take(get_validator_churn_limit(state, context) as usize)
+    {
+        let validator = &mut state.validators[*i];
+        validator.activation_epoch = compute_activation_exit_epoch(current_epoch, context);
+    }
 
     Ok(())
 }
--- a/src/phase0/state_transition/epoch_processing.rs
+++ b/src/phase0/state_transition/epoch_processing.rs
@@ -179,23 +179,29 @@ pub fn process_registry_updates<
         .validators
         .iter()
         .enumerate()
-        .filter(|&(_, validator)| is_eligible_for_activation(state, validator))
-        .map(|(i, _)| i)
+        .filter_map(|(index, validator)| {
+            if is_eligible_for_activation(state, validator) {
+                Some(index)
+            } else {
+                None
+            }
+        })
         .collect::<Vec<ValidatorIndex>>();
     // Order by the sequence of activation_eligibility_epoch setting and then index
-    activation_queue.sort_by(|i, j| {
-        let a = &state.validators[*i];
-        let b = &state.validators[*j];
+    activation_queue.sort_by(|&i, &j| {
+        let a = &state.validators[i];
+        let b = &state.validators[j];
         (a.activation_eligibility_epoch, i).cmp(&(b.activation_eligibility_epoch, j))
     });
 
     // Dequeued validators for activation up to churn limit
+    let activation_exit_epoch = compute_activation_exit_epoch(current_epoch, context);
     for i in activation_queue
-        .iter()
-        .take(get_validator_churn_limit(state, context) as usize)
+        .into_iter()
+        .take(get_validator_churn_limit(state, context))
     {
-        let validator = &mut state.validators[*i];
-        validator.activation_epoch = compute_activation_exit_epoch(current_epoch, context);
+        let validator = &mut state.validators[i];
+        validator.activation_epoch = activation_exit_epoch;
     }
 
     Ok(())
--- a/src/phase0/state_transition/epoch_processing.rs
+++ b/src/phase0/state_transition/epoch_processing.rs
@@ -1,15 +1,14 @@
 use crate::phase0::beacon_state::{BeaconState, HistoricalBatchAccumulator};
 use crate::phase0::operations::PendingAttestation;
 use crate::phase0::state_transition::{
-    compute_activation_exit_epoch, decrease_balance, get_block_root, get_current_epoch,
-    get_previous_epoch, get_randao_mix, get_total_active_balance, get_validator_churn_limit,
-    initiate_validator_exit, is_active_validator, is_eligible_for_activation,
-    is_eligible_for_activation_queue, Context, Error,
+    decrease_balance, get_block_root, get_current_epoch, get_previous_epoch, get_randao_mix,
+    get_total_active_balance, initiate_validator_exit, is_active_validator,
+    is_eligible_for_activation, is_eligible_for_activation_queue, Context, Error,
 };
 use crate::phase0::validator::Validator;
 use crate::primitives::{Epoch, Gwei, ValidatorIndex};
 use ssz_rs::prelude::*;
-use std::mem;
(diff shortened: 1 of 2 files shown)


src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,63 @@ pub fn process_registry_updates<
...
+
+        if is_active_validator(validator, current_epoch)
+            && validator.effective_balance <= context.ejection_balance
+        {
+            initiate_validator_exit(state, i, context);
+        }
+    }
+
+    // Queue validators eligible for activation and not yet dequeued for activation
+    let _activation_queue = state
```
buchhlz2, line 178: Here's where step 2 begins & issues begin -- and the type on `_activation_queue` is `()` with the current code. If I remove the `sort_by`, the type is a `Vec<(usize, &Validator)>`, which is what I'd expect. So, I'm thinking it's an issue with my handling of `sort_by` but am not entirely sure here.

buchhlz2, line 178: Here's where step 2 begins & issues begin -- and the type on `_activation_queue` is `()` with the current code. If I remove the `sort_by`, the type is a `Vec<(usize, &Validator)>`, which is what I'd expect. So, I'm thinking it's an issue with my handling of `sort_by` but am not entirely sure here.

buchhlz2, line 178: Here's where step 2 begins & issues begin -- and the type on `_activation_queue` is `()` with the current code. If I remove the `sort_by`, the type is a `Vec<(usize, &Validator)>`, which is what I'd expect. So, I'm thinking it's an issue with my handling of `sort_by` but am not entirely sure here.

ralexstokes: yeah, i think this is just a chaining issue...

you are calling `collect` to materialize the `Vec` and then immediately calling `sort_by` which mutates the Vec in-place.

but you are capturing the result of this (just the `()` as the sort mutates the vec in place and simply returns `()` here) as `activation_queue` so everything seems as expected.


if i'm understanding correctly, you just need to say `let mut activation_queue = state./*... elided ...*/.collect();` and then on the next line `activation_queue.sort_by(...)`

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,63 @@ pub fn process_registry_updates<
...
+        if is_eligible_for_activation_queue(validator, context) {
+            validator.activation_eligibility_epoch = current_epoch + 1;
+        }
+
+        if is_active_validator(validator, current_epoch)
+            && validator.effective_balance <= context.ejection_balance
+        {
+            initiate_validator_exit(state, i, context);
+        }
+    }
```
ralexstokes, line 175: this part looks good

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,63 @@ pub fn process_registry_updates<
...
+        .filter(|&(_, validator)| is_eligible_for_activation(state, validator))
+        .collect::<Vec<(ValidatorIndex, &Validator)>>()
+        // Order by the sequence of activation_eligibility_epoch setting and then index
+        .sort_by(|a, b| {
+            let a_activation_eligibility_epoch = a.1.activation_eligibility_epoch;
+            let a_index = a.0;
+            let b_activation_eligibility_epoch = b.1.activation_eligibility_epoch;
+            let b_index = b.0;
+            match a_activation_eligibility_epoch.cmp(&b_activation_eligibility_epoch) {
+                cmp::Ordering::Less => cmp::Ordering::Less,
```
ralexstokes, line 191: random aside but when i run into needing to match on the variants of `Ordering` i just import `std::cmp::Ordering` so that you can drop all of these `cmp`

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,63 @@ pub fn process_registry_updates<
...
+                cmp::Ordering::Greater => cmp::Ordering::Greater,
+                cmp::Ordering::Equal => match a_index.cmp(&b_index) {
+                    cmp::Ordering::Less => cmp::Ordering::Less,
+                    cmp::Ordering::Greater => cmp::Ordering::Greater,
+                    // TODO: throw some error if the valators' indices are equal -- should not be possible
+                    cmp::Ordering::Equal => cmp::Ordering::Equal,
+                },
+            }
```
ralexstokes, line 199: so i think we can simplify all of this by taking advantage of the existing impl of `PartialOrd` for tuples...

something like:

```rust
.sort_by(|(a_index, a), (b_index, b)| { (a.epoch, a_index) < (b.epoch, b_index) })
```

where you can use tuple destructuring in the closure argument to save some typing

ralexstokes (commented):
this definitely looks like the right direction!

i left some suggestions below, and would also consider only working w/ validator indices as you move to the final activations in this code

this means something like:

iterating over the enumerated validators (like you have now)
filtering for the eligible ones
but only `collect`ing a `Vec<ValidatorIndex>` (and i'll call your attention to [`filter_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map) )

then, when you go to sort, you can just load the ref to the validator (like you did for the first step)

```rust
...sort_by(|i, j| { let a = &state.validators[i]; let b = &state.validators[j]; (a.epoch, i).cmp(&(b.epoch, j)) });
```
where i *think* that won't lead to borrow checker issues...


then you should be able to use the rest of your code:

`for i in queue.iter.take(...) { let validator = &mut state.validators[i]; /* do the mutation */ }`

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Kk yeah good call

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Ah, yeah, that was straightforward...have to have the `activation_queue` created to be able to sort it.

buchhlz2:
A couple things to note:
- I opted for a plain `filter` then `map` instead a `filter_map` -- I would have had to add a `match` on `is_eligible_for_activation` to then map the values accordingly (e.g., `true => Some(i), false => None`), so I felt `filter` then `map` was more concise (unless there's a performance reason or something that prefers `filter_map`...I don't have a strong preference)
- I had to dereference `i` and `j` in the `sort_by` -- haven't really had to use `*` before so calling this out just in case

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,48 @@ pub fn process_registry_updates<
...
+    let mut activation_queue = state
+        .validators
+        .iter()
+        .enumerate()
+        .filter(|&(_, validator)| is_eligible_for_activation(state, validator))
+        .map(|(i, _)| i)
+        .collect::<Vec<ValidatorIndex>>();
+    // Order by the sequence of activation_eligibility_epoch setting and then index
+    activation_queue.sort_by(|i, j| {
+        let a = &state.validators[*i];
```
buchhlz2, line 187: Dereference here

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,48 @@ pub fn process_registry_updates<
...
+            initiate_validator_exit(state, i, context);
+        }
+    }
+
+    // Queue validators eligible for activation and not yet dequeued for activation
+    let mut activation_queue = state
+        .validators
+        .iter()
+        .enumerate()
+        .filter(|&(_, validator)| is_eligible_for_activation(state, validator))
```
buchhlz2, line 182: `filter` then `map` here

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,48 @@ pub fn process_registry_updates<
...
+    activation_queue.sort_by(|i, j| {
+        let a = &state.validators[*i];
+        let b = &state.validators[*j];
+        (a.activation_eligibility_epoch, i).cmp(&(b.activation_eligibility_epoch, j))
+    });
+
+    // Dequeued validators for activation up to churn limit
+    for i in activation_queue
+        .iter()
+        .take(get_validator_churn_limit(state, context) as usize)
```
ralexstokes, line 195: could you fold `as usize` into `get_validator_churn_limit` (and change the return type) and then remove it from here?

it is more "rusty" to have this value be `usize` and we will just end up casting it in a few other places where this function will be called

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,48 @@ pub fn process_registry_updates<
...
+    // Order by the sequence of activation_eligibility_epoch setting and then index
+    activation_queue.sort_by(|i, j| {
+        let a = &state.validators[*i];
+        let b = &state.validators[*j];
+        (a.activation_eligibility_epoch, i).cmp(&(b.activation_eligibility_epoch, j))
+    });
+
+    // Dequeued validators for activation up to churn limit
+    for i in activation_queue
+        .iter()
```
ralexstokes, line 194: i would actually do this and just take the indices by value
```suggestion
        .into_iter()
```

the idea here is that you will already be shuffling say 64 bits around, and taking them by reference mean you will need to deref the values to get the 64 bits you need

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,48 @@ pub fn process_registry_updates<
...
+        .take(get_validator_churn_limit(state, context) as usize)
+    {
+        let validator = &mut state.validators[*i];
```
ralexstokes, line 197: then: 
```suggestion
        let validator = &mut state.validators[i];
```

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,48 @@ pub fn process_registry_updates<
...
+        validator.activation_epoch = compute_activation_exit_epoch(current_epoch, context);
```
ralexstokes, line 198: possibly splitting hairs but you could hoist `compute_activation_exit_epoch` out of this loop so you only compute the value once, and then use that value to write into the validator

src/phase0/state_transition/epoch_processing.rs
ralexstokes: this is fine -- one thing that is a bit more idiomatic is to use ref destructuring:

```rust
.sort_by(|&i, &j| { ... })
```

src/phase0/state_transition/epoch_processing.rs
ralexstokes: this isn't the end of the world but it does extend the iterator chain so there is some performance hit...

the idea with filter map:

```rust
.filter_map(|&(index, validator)| {
    if is_eligible(validator) {
        Some(index)
    } else {
        None
    }
})
```

but not a huge deal

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,48 @@ pub fn process_registry_updates<
...
+        .iter()
+        .enumerate()
+        .filter(|&(_, validator)| is_eligible_for_activation(state, validator))
+        .map(|(i, _)| i)
+        .collect::<Vec<ValidatorIndex>>();
+    // Order by the sequence of activation_eligibility_epoch setting and then index
+    activation_queue.sort_by(|i, j| {
+        let a = &state.validators[*i];
+        let b = &state.validators[*j];
+        (a.activation_eligibility_epoch, i).cmp(&(b.activation_eligibility_epoch, j))
```
ralexstokes, line 189: aside: i'm wondering now if it is possible for `i > j` *ever*

perhaps if the sort wasn't stable, but the pyspec `sort` should be....

ralexstokes (commented):
left a few minor suggestions towards more imo idiomatic style

close to being ready for merge!

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Ah, right, change makes sense then -- done

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Good call, missed that

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Not sure if I fully did what you were thinking here -- only change was making it `.sort_by(|&i, &j| { ... })`

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Hmm like, will the `ValidatorIndex` in `state.validators` ever not be in sequential order? I haven't really thought through how validators are ordered once initially added to state, at least in a while.

Or, just a callout on rust's `sort_by` and edge cases where `i` could be `> j` for some reason?

ralexstokes: yeah i meant aside from this rust code -- like in the pyspec, we break ties by validator index but i'm not sure it actually ever matters.... i guess it is not worth changing now though :)

src/phase0/state_transition/epoch_processing.rs
ralexstokes: yeah and then you dropped the manual derefs `*` -- this is what i meant!

ralexstokes (approved):
updates are excellent!

it seems like somehow history that was already on main has been put on this branch...

if you rebase this branch against main then we can merge it in!

buchhlz2:
Hmm yeah idk exactly what I did, but I think I tried to merge the latest changes from upstream into this branch...couldn't figure out the drops in `rebase -i` so followed something on here: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

Aka lmk if it's still a mess, and I could maybe just open a new PR

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,54 @@ pub fn process_registry_updates<
...
+        .iter()
+        .enumerate()
+        .filter_map(|(index, validator)| {
+            if is_eligible_for_activation(state, validator) {
+                Some(index)
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<ValidatorIndex>>();
```
ralexstokes, line 189: nitpick: 
```suggestion
        .collect::<Vec<ValidatorIndex>>();
        
```

ralexstokes (approved):
looks great!

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -154,8 +156,55 @@ pub fn process_registry_updates<
...
+
```
ralexstokes, line 190: whitespace fix lol

```suggestion

```

PR #46, opened 2022-01-07
feat: add `process_rewards_and_penalties` impl & helper fns

buchhlz2:
Add skeletons for all remaining functions yet to be implemented but required by `process_rewards_and_penalties` and its related helpers/subfunctions (where duplicate fns are listed here to define their missing dependencies):
- `process_rewards_and_penalties`
    - `get_attestation_deltas`
        - `get_attestation_component_deltas`
            - `get_unslashed_attesting_indices`
            - `get_eligible_validator_indices`
            - `is_in_inactivity_leak`
                - `get_finality_delay`
            - `get_base_reward`
        - `get_source_deltas`
        - `get_target_deltas`
        - `get_head_deltas`
            - `get_matching_head_attestations`
        - `get_inclusion_delay_deltas`
            - `get_unslashed_attesting_indices`
            - `get_proposer_reward`
            - `get_base_reward`
            - `get_proposer_reward`
        - `get_inactivity_penalty_deltas`
            - `is_in_inactivity_leak`
                - `get_finality_delay`
            - `get_unslashed_attesting_indices`
            - `get_eligible_validator_indices`
            - `get_base_reward`
            - `get_proposer_reward`
            - `get_finality_delay`

### DIFF
--- a/src/phase0/state_transition/epoch_processing.rs
+++ b/src/phase0/state_transition/epoch_processing.rs
@@ -1,12 +1,14 @@
 use crate::phase0::beacon_state::{BeaconState, HistoricalBatchAccumulator};
 use crate::phase0::operations::PendingAttestation;
 use crate::phase0::state_transition::{
-    compute_activation_exit_epoch, decrease_balance, get_block_root, get_current_epoch,
-    get_previous_epoch, get_randao_mix, get_total_active_balance, get_validator_churn_limit,
-    initiate_validator_exit, is_active_validator, is_eligible_for_activation,
-    is_eligible_for_activation_queue, Context, Error,
+    compute_activation_exit_epoch, decrease_balance, get_attesting_indices, get_block_root,
+    get_current_epoch, get_previous_epoch, get_randao_mix, get_total_active_balance,
+    get_total_balance, get_validator_churn_limit, initiate_validator_exit, is_active_validator,
+    is_eligible_for_activation, is_eligible_for_activation_queue, Context, Error, HashSet,
 };
+use crate::phase0::BASE_REWARDS_PER_EPOCH;
 use crate::primitives::{Epoch, Gwei, ValidatorIndex};
+use integer_sqrt::IntegerSquareRoot;
 use ssz_rs::prelude::*;
 use std::mem;
 
@@ -84,6 +86,69 @@ pub fn get_matching_target_attestations<
     Ok(source_attestations
         .iter()
         .filter(move |a| a.data.target.root == *block_root))
+}
+
+pub fn _get_matching_head_attestations<
+    const SLOTS_PER_HISTORICAL_ROOT: usize,
+    const HISTORICAL_ROOTS_LIMIT: usize,
+    const ETH1_DATA_VOTES_BOUND: usize,
+    const VALIDATOR_REGISTRY_LIMIT: usize,
+    const EPOCHS_PER_HISTORICAL_VECTOR: usize,
+    const EPOCHS_PER_SLASHINGS_VECTOR: usize,
+    const MAX_VALIDATORS_PER_COMMITTEE: usize,
+    const PENDING_ATTESTATIONS_BOUND: usize,
+>(
+    _state: &BeaconState<
+        SLOTS_PER_HISTORICAL_ROOT,
+        HISTORICAL_ROOTS_LIMIT,
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    _context: &Context,
+) -> Result<(), Error> {
+    Ok(())
+}
+
+pub fn get_unslashed_attesting_indices<
+    const SLOTS_PER_HISTORICAL_ROOT: usize,
+    const HISTORICAL_ROOTS_LIMIT: usize,
+    const ETH1_DATA_VOTES_BOUND: usize,
+    const VALIDATOR_REGISTRY_LIMIT: usize,
+    const EPOCHS_PER_HISTORICAL_VECTOR: usize,
+    const EPOCHS_PER_SLASHINGS_VECTOR: usize,
+    const MAX_VALIDATORS_PER_COMMITTEE: usize,
+    const PENDING_ATTESTATIONS_BOUND: usize,
+>(
+    state: &BeaconState<
+        SLOTS_PER_HISTORICAL_ROOT,
+        HISTORICAL_ROOTS_LIMIT,
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    attestations: &[PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>],
+    context: &Context,
+) -> Result<HashSet<ValidatorIndex>, Error> {
+    let output = HashSet::new();
+    for a in attestations.iter() {
+        output.union(&get_attesting_indices(
+            state,
+            &a.data,
+            &a.aggregation_bits,
+            context,
+        )?);
+    }
+    Ok(output
+        .into_iter()
+        .filter(|&i| !state.validators[i].slashed)
+        .collect::<HashSet<ValidatorIndex>>())
 }
 
 pub fn process_justification_and_finalization<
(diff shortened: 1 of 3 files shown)


buchhlz2:
1) Since there's a lot to unpack here, should I create individual PRs for these fns that are must be implemented? i.e., open a separate issue for `process_rewards_and_penalties`?
2) I might take a stab at `process_justification_and_finalization` before tackling this one, so I might just want to just add the skeletons now and come back later -- not sure if where I put everything is correct (`mod.rs` vs. `epoch_processing.rs`)

ralexstokes:
> should I create individual PRs for these fns

up to you about PRs; you don't need to make a new issue, we can just link everything in #29.

this could be a PR for the "helper" fns and then perhaps one PR per reward type

ralexstokes:
> stab at process_justification_and_finalization before tackling this one

taht works!

> not sure if where I put everything is correct (mod.rs vs. epoch_processing.rs)

i would put all of this stuff into `epoch_processing.rs`

buchhlz2:
For any of these that return a value, is using `Result` desired or just the naked value? E.g., `is_in_inactivity_leak` returns `bool` -- looking in the `gensis.rs` file, I see an example like `is_in_inactivity_leak` returns just the value, but other fns are wrapping the return in a `Result`.

So, is there a rule of thumb I should go by?

ralexstokes:
> So, is there a rule of thumb I should go by?

main question is if there is possibility for an error or not

if something, like `is_in_activity_leak` returns just a flat `bool` and cannot error then by all means, just return the `bool`.

i think everywhere we have a `Result` it is bc somewhere in the call stack you can have a failure

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    attestations: &[PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>],
+    context: &Context,
+) -> Result<(Vec<Gwei>, Vec<Gwei>), Error> {
+    // Helper with shared logic for use by get source, target, and head deltas functions
```
buchhlz2, line 680: Just to make sure I'm consistent -- preference on including the spec's inline comments? Keep or nah?

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+    let mut rewards = vec![0, state.validators.len() as Gwei];
+    let mut penalties = vec![0, state.validators.len() as Gwei];
+    let total_balance = get_total_active_balance(state, context)?;
+    let unslashed_attesting_indices =
+        get_unslashed_attesting_indices(state, attestations, context)?;
+    let attesting_balance = get_total_balance(state, &unslashed_attesting_indices, context)?;
```
buchhlz2, line 686: This was the reason why I changed `get_total_balance` in `mod.rs`, so I'm open to suggestions if that's not desired. Basically, I was getting the error `borrow of moved value: unslashed_attesting_indices` -- which (checking my understanding) means `unslashed_attesting_indices` was being moved to `attesting_balance` when it calls `get_total_balance`, so the compiler can't access the original variable anymore. Thus, I couldn't access `unslashed_attesting_indices` within the `for` loop

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+            } else {
+                let reward_numerator =
+                    get_base_reward(state, i, context)? * (attesting_balance / increment);
+                rewards[i] += reward_numerator / (total_balance / increment);
+            }
+        } else {
+            penalties[i] += get_base_reward(state, i, context)?;
+        }
+    }
+    Ok((rewards, penalties))
```
buchhlz2, line 703: Also, random formatting preference -- return `Ok()`s with whitespace above it, or remove it (as it has been here)?

buchhlz2:
Note -- I pushed this thinking I was at the point of wanting to merge what I had for some reason...but there are a few helpers left. And at that point, I'd only be a few fn impls away, so I'm thinking I'll just finish it all in this one PR.

src/phase0/state_transition/epoch_processing.rs
ralexstokes: your call, i don't mind them

src/phase0/state_transition/epoch_processing.rs
ralexstokes: likely context dependent :)

everyone will have their own preferences on formatting / layout -- i'd say just pick which one looks better to you.

my personal rule of thumb has to do w/ how related the return value is to the code before it -- if things are linked conceptually then i would group them closer together -- if the prior chunk of code is not for some reason then i'd put in some whitespace.... but not something to get super hung up on imo

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    index: ValidatorIndex,
+    context: &Context,
+) -> Result<Gwei, Error> {
+    let total_balance = get_total_active_balance(state, context)?;
+    let effective_balance = state.validators[index].effective_balance;
+    Ok(effective_balance * context.base_reward_factor
+        / total_balance.integer_sqrt()
+        / BASE_REWARDS_PER_EPOCH as u64)
```
ralexstokes, line 539: pretty sure i typed `BASE_REWARDS_PER_EPOCH` as `usize` but given its usage, it may be better to just define as `u64` and avoid these casts

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+        HISTORICAL_ROOTS_LIMIT,
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    context: &Context,
+) -> u64 {
```
ralexstokes, line 589: i would just go ahead and type this as `Epoch`
```suggestion
) -> Epoch {
```

it doesn't strictly align as the spec "puns" it a bit already

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+        .filter_map(|(i, v)| {
+            if is_active_validator(v, previous_epoch)
+                || (v.slashed && previous_epoch + 1 < v.withdrawable_epoch)
+            {
+                Some(i)
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<ValidatorIndex>>()
```
ralexstokes, line 654: so one thing we can consider here is to avoid the allocation with the `Vec`....

perhaps it will ultimately not matter but i have been leaning towards using "existential" types for this with `impl Trait`:

```rust
fn get_eligible_validator_indices(...) -> impl Iterator<Item=ValidatorIndex> { ... }
```


and then you can drop the final `collect` you have here.

then the caller can decide what to do

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    attestations: &[PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>],
+    context: &Context,
+) -> Result<(Vec<Gwei>, Vec<Gwei>), Error> {
+    // Helper with shared logic for use by get source, target, and head deltas functions
+    let mut rewards = vec![0, state.validators.len() as Gwei];
```
ralexstokes, line 681: minor but you could pull out `state.validators.len()` into one `let validator_count = ...;`

and then not call this function twice.

ralexstokes, line 681: you are also looking for a slightly diff syntax:

```suggestion
    let mut rewards = vec![0; state.validators.len()];
```

you make a vector of two elements, when what you want is a vector of N element of the default value

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+    let mut penalties = vec![0, state.validators.len() as Gwei];
```
ralexstokes, line 682: ditto on this init form

src/phase0/state_transition/epoch_processing.rs
ralexstokes: yeah, understanding is correct... i'll look at the change, prolly the right thing to do

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +509,350 @@ pub fn process_participation_record_updates<
...
+    let total_balance = get_total_active_balance(state, context)?;
+    let unslashed_attesting_indices =
+        get_unslashed_attesting_indices(state, attestations, context)?;
+    let attesting_balance = get_total_balance(state, &unslashed_attesting_indices, context)?;
+    for i in get_eligible_validator_indices(state, context) {
+        if unslashed_attesting_indices.contains(&i) {
+            let increment = context.effective_balance_increment; // Factored out from balance totals to avoid uint64 overflow
```
ralexstokes, line 689: could pull out of this loop structure, to avoid having to define each time

src/phase0/state_transition/mod.rs
```diff
@@ -975,7 +975,7 @@ pub fn get_total_balance<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    indices: HashSet<ValidatorIndex>,
+    indices: &HashSet<ValidatorIndex>,
```
ralexstokes, line 978: this would work....

another option is to "erase" the specific type with the existential typing i mentioned above...

```suggestion
    indices: impl Iterator<Item=ValidatorIndex>,
```

and then below you could do `HashSet::from_iter(...).iter()` (which we do bc we want to ensure set semantics at that call site)

and then up above you could do something like:

```rust
let attesting_balance = get_total_balance(state, unslashed_attesting_indices.iter(), context)?;
```

which is functionally the same thing....

.... after typing it out, it seems fine enough to just keep it how you have it. it follows the spec and i don't see any crazy performance differential -- if we accepted more types of inputs to this function or used it in more places in more ways this could change but this PR introduces the only other use of `get_total_balance`

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -86,6 +88,69 @@ pub fn get_matching_target_attestations<
...
+            state,
+            &a.data,
+            &a.aggregation_bits,
+            context,
+        )?);
+    }
+    Ok(output
+        .into_iter()
+        .filter(|&i| !state.validators[i].slashed)
+        .collect::<HashSet<ValidatorIndex>>())
```
ralexstokes, line 151: this is another place we could consider removing the intermediate allocations by dropping `collect` and considering how we build the set `output` but i think for now this is fine -- there are bigger hot spots we will hit and i don't know if we will hit this level of optimization in this project tbh

ralexstokes (commented):
excellent! i left some comments

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Cool...I'll leave this one alone for now.

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Having a little trouble getting the `impl Iterator` working -- trying to use lifetime values (like in `get_matching_target_attestations`), and I can't figure out how to align the return value. I'm getting an error:
```
expected enum `Option<&'a usize>`
found enum `Option<usize>`
```
...namely, idk how to get `usize` to be `&'a usize` 
https://gist.github.com/buchhlz2/9028b4387c10b30592edfefcb3d8e379

ralexstokes: i would just drop the ref... if you say `Item=ValidatorIndex` it will prolly work

i think the compiler is just doing a copy for you -- i don't really see a reason to work w/ ref's here

buchhlz2: Ah...I couldn't figure out how to use the lifetime unless there was a `&` involved....but yeah that works with `impl Iterator<Item = ValidatorIndex> + 'a` along with a `move` in the closure

ralexstokes: ah fair point...

i think what i am *trying* to do would actually imply a `cloned` after the `filter_map` and without making this a `move` closure...

if this is becoming a time sink, its fine to just `collect` into a `Vec` imo

src/phase0/state_transition/epoch_processing.rs
buchhlz2: If needed -- this is a very minor change with the latest commit. Like <5 mins of work, so I'll default to you if we want an "existential" type here to keep things consistent

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Oops, just saw this. Well, I opted for the `impl Iterator` and a closure `move` so lmk if I should refactor.

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -86,6 +90,74 @@ pub fn get_matching_target_attestations<
...
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    epoch: Epoch,
+    context: &Context,
+) -> Result<impl Iterator<Item = &'a PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>>, Error> {
+    let matching_target_attestations = get_matching_target_attestations(state, epoch, context)?;
+    Ok(matching_target_attestations.filter(|a| {
+        a.data.beacon_block_root == *get_block_root_at_slot(state, a.data.slot).unwrap()
```
buchhlz2, line 119: I had to `unwrap()` here -- just checking if this is the desired way to handle things or not. If not, lmk -- I also do this in `get_inclusion_delay_deltas` in 3 different spots.

My understanding is that I'd have to convert this to a `for` loop (or some other methods like `try_into`) to properly use `?` instead of `unwrap()` due this occurring in a closure.

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +526,437 @@ pub fn process_participation_record_updates<
...
+    // Return attestation reward/penalty deltas for each validator.
+    let (source_rewards, source_penalties) = get_source_deltas(state, context)?;
+    let (target_rewards, target_penalties) = get_target_deltas(state, context)?;
+    let (head_rewards, head_penalties) = get_head_deltas(state, context)?;
+    let (inclusion_delay_rewards, _) = get_inclusion_delay_deltas(state, context)?;
+    let (_, inactivity_penalties) = get_inactivity_penalty_deltas(state, context)?;
+
+    let validator_count = state.validators.len();
+    let mut rewards = vec![0; validator_count];
+    for i in 0..validator_count {
```
buchhlz2, line 947: Wasn't sure if this was the best approach or doing:
- `rewards.map(...)` instead of the `for` loop
- Making `rewards` an empty `Vec` and then pushing into it during the `for` loop

Same obv for below

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +526,437 @@ pub fn process_participation_record_updates<
...
+    for i in get_unslashed_attesting_indices(state, matching_source_attestations, context)? {
+        let attestation = matching_source_attestations
+            .iter()
+            .filter(|a| {
+                get_attesting_indices(state, &a.data, &a.aggregation_bits, context)
+                    .unwrap()
+                    .contains(&i)
+            })
+            .min_by(|a, b| a.inclusion_delay.cmp(&b.inclusion_delay))
+            .unwrap();
```
ralexstokes, line 852: ditto on `unwrap`

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +526,437 @@ pub fn process_participation_record_updates<
...
+    >,
+    context: &Context,
+) -> Result<(Vec<Gwei>, Vec<Gwei>), Error> {
+    // Return attester micro-rewards/penalties for target-vote for each validator.
+    let previous_epoch = get_previous_epoch(state, context);
+    let matching_target_attestations =
+        get_matching_target_attestations(state, previous_epoch, context)?
+            .cloned()
+            .collect::<Vec<_>>();
+    get_attestation_component_deltas(state, matching_target_attestations.as_slice(), context)
```
ralexstokes, line 781: so the types may get a little wonky, but ideally we avoid copying the `PendingAttestations`

they are owned by the `BeaconState` and (again ideally) we only pass in references to the right ones to `get_attestation_component_deltas`

i'd suggest doing the simple thing first to get it right, we can copy/clone/collect into `Vec` (or whatever is needed) to write the code and then can consider ways to avoid the copies where possible

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +526,437 @@ pub fn process_participation_record_updates<
...
+    >,
+    context: &Context,
+) -> Result<(Vec<Gwei>, Vec<Gwei>), Error> {
+    // Return attester micro-rewards/penalties for head-vote for each validator.
+    let previous_epoch = get_previous_epoch(state, context);
+    let matching_head_attestations =
+        get_matching_head_attestations(state, previous_epoch, context)?
+            .cloned()
+            .collect::<Vec<_>>();
+    get_attestation_component_deltas(state, matching_head_attestations.as_slice(), context)
```
ralexstokes, line 812: random nitpick, but as far as i understand, it is a bit more idiomatic to use `&t` for `t: Vec<T>` rather than call the method `as_slice`

src/phase0/state_transition/epoch_processing.rs
ralexstokes: whatever gets correct code written in the shortest possible time :)

one thing to consider is that we will likely come optimize this and a big source of speed ups will be eliminating the "iter over all validators" pattern that ends up everywhere in the "naive" spec

ralexstokes (commented):
great updates!

main remaining work will be massaging the error handling to drop all of the `unwrap`s

and it is also worth considering how we can skip some of the copying when tallying attestations

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Agreed, idk why I used that hah

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +530,433 @@ pub fn process_participation_record_updates<
...
+        let mut attestations = Vec::new();
+        for a in &matching_source_attestations {
+            if get_attesting_indices(state, &a.data, &a.aggregation_bits, context)?.contains(&i) {
+                attestations.push(*a)
+            }
+        }
+        let attestation = attestations
+            .iter()
+            .min_by(|&a, &b| a.inclusion_delay.cmp(&b.inclusion_delay))
+            .ok_or(Error::InvalidAttestation)?;
```
buchhlz2, line 857: I wasn't sure what's ideal for error handling an `Option`, so I opted for `ok_or` instead of doing some `match`. I also wasn't sure exactly what to use -- sort of inserted `InvalidAttestation` as it was already an existing type, but I'm not sure if it's what it's intended for.

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -86,6 +90,78 @@ pub fn get_matching_target_attestations<
...
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    attestations: &[&PendingAttestation<MAX_VALIDATORS_PER_COMMITTEE>],
+    context: &Context,
+) -> Result<HashSet<ValidatorIndex>, Error> {
+    let output = HashSet::new();
+    for a in attestations.iter() {
+        output.union(&get_attesting_indices(
```
buchhlz2, line 152: So my clippy didn't catch this, but the `union` isn't mutating the `output` (which I should mark as `mut`)...resolving...

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -745,9 +749,11 @@ pub fn get_source_deltas<
 ) -> Result<(Vec<Gwei>, Vec<Gwei>), Error> {
     // Return attester micro-rewards/penalties for source-vote for each validator.
     let previous_epoch = get_previous_epoch(state, context);
-    let matching_source_attestations =
-        get_matching_source_attestations(state, previous_epoch, context)?;
-    get_attestation_component_deltas(state, matching_source_attestations, context)
+    let mut matching_source_attestations = Vec::new();
+    for a in get_matching_source_attestations(state, previous_epoch, context)?.iter() {
+        matching_source_attestations.push(a);
+    }
```
ralexstokes, line 755: a bit easier to see what is going on w/ `collect` imo

`get_matching_source_attestations(...)?.iter().collect::<Vec<_>>()`

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -838,19 +840,22 @@ pub fn get_inclusion_delay_deltas<
...
     let validator_count = state.validators.len();
     let mut rewards = vec![0; validator_count];
-    let matching_source_attestations =
-        get_matching_source_attestations(state, previous_epoch, context)?;
-    for i in get_unslashed_attesting_indices(state, matching_source_attestations, context)? {
-        let attestation = matching_source_attestations
+    let mut matching_source_attestations = Vec::new();
+    for a in get_matching_source_attestations(state, previous_epoch, context)?.iter() {
+        matching_source_attestations.push(a);
+    }
```
ralexstokes, line 846: same w/ `collect` rather than `iter` and `push`

src/phase0/state_transition/epoch_processing.rs
ralexstokes: so we can consider what kind of error this is and use that to inform what the `Option` here really *means*

and we can see that `min_by` returns `None` only when `attestations` would be empty, and we know that `attestations` will never be empty as our "source" material is ultimately from the `matching_source_attestations`

so in this case, we can actually just `unwrap` the option -- and in fact i'd suggest using `expect` with a message like "at least one attestation in collection" (as it provides helpful context if something does blow up unexpectedly)

src/phase0/state_transition/mod.rs
```diff
@@ -46,6 +46,8 @@ pub enum Error {
     },
     #[error("overflow")]
     Overflow,
+    #[error("invalid attestation")]
+    InvalidAttestation,
```
ralexstokes, line 50: see above comment, i think we can drop this -- if you did need this here, i would want to consider it amongst the other variants as `InvalidAttestation` really just refers to an invalid att in the context of a bad block, not some error during epoch processing

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -888,14 +893,9 @@ pub fn get_inactivity_penalty_deltas<
...
         let matching_target_attestations =
-            get_matching_target_attestations(state, previous_epoch, context)?
-                .cloned()
-                .collect::<Vec<_>>();
-        let matching_target_attesting_indices = get_unslashed_attesting_indices(
-            state,
-            matching_target_attestations.as_slice(),
-            context,
-        )?;
+            get_matching_target_attestations(state, previous_epoch, context)?.collect::<Vec<_>>();
```
ralexstokes, line 896: i may be sending you down the wrong rabbit hole but we have this patter now of "call get_matching_*_atts" and then `collect` them into a `Vec`

just so we can pipe something into `get_unslashed_attesting_indices` -- i have wonder if we can skip the `collect` by relaxing `get_unslashed_attesting_indices` so that it accepts anything that `impl Iterator<Item=&PendingAtt>`?

buchhlz2: Yeah, I think that logic makes sense.

So, if `get_unslashed_attesting_indices` is relaxed, then it probably makes sense to also relax `get_attestation_component_deltas` since `attestations` are simply passed from the `get_attestation_component_deltas` `attestations` param to `get_unslashed_attesting_indices`.

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -144,17 +145,20 @@ pub fn get_unslashed_attesting_indices<
...
+    for a in attestations {
+        output = output
+            .union(&get_attesting_indices(
+                state,
+                &a.data,
+                &a.aggregation_bits,
+                context,
+            )?)
+            .cloned()
+            .collect::<HashSet<_>>();
```
ralexstokes, line 161: suggestion to skip a bunch of intermittent allocations:

```rust
let mut output = HashSet::new();
for a in attestations {
    for index in get_attesting_indices(...)? {
        if !state.validators[index].slashed {
            output.insert(index);
        }
    }
}
Ok(output)
```

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +527,430 @@ pub fn process_participation_record_updates<
...
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    context: &Context,
+) -> Result<(Vec<Gwei>, Vec<Gwei>), Error> {
+    // Return attestation reward/penalty deltas for each validator.
+    let (source_rewards, source_penalties) = get_source_deltas(state, context)?;
+    let (target_rewards, target_penalties) = get_target_deltas(state, context)?;
+    let (head_rewards, head_penalties) = get_head_deltas(state, context)?;
+    let (inclusion_delay_rewards, _) = get_inclusion_delay_deltas(state, context)?;
+    let (_, inactivity_penalties) = get_inactivity_penalty_deltas(state, context)?;
```
ralexstokes, line 937: may as well just drop the `vec!` that just immediately get discarded

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -444,6 +527,430 @@ pub fn process_participation_record_updates<
...
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    context: &Context,
+) -> Result<(Vec<Gwei>, Vec<Gwei>), Error> {
+    // Return attester micro-rewards/penalties for source-vote for each validator.
+    let previous_epoch = get_previous_epoch(state, context);
+    let matching_source_attestations =
+        get_matching_source_attestations(state, previous_epoch, context)?.iter();
+    get_attestation_component_deltas(state, matching_source_attestations, context)
```
ralexstokes, line 752: a bit minor but you did this the other way in other places and i have a preference for that as the thing that is bound by a name is (more) concrete data, rather than something like an "Iter"
```suggestion
    let matching_source_attestations =
        get_matching_source_attestations(state, previous_epoch, context)?;
    get_attestation_component_deltas(state, matching_source_attestations.into_iter(), context)
```

and i think you can do `into_iter` over just `iter` here

buchhlz2: Good catch -- I can definitely at least put the `iter()` within the fn call. However, using `matching_source_attestations.into_iter()` brings a couple of errors:
1. `get_attestation_component_deltas` expects a reference, but `into_iter` makes it a value -- i.e., I get an error `expected reference &PendingAttestation, found struct PendingAttestation`
2. Thus, if I use `&matching_source_attestations.into_iter()`, then I run into iterator issues: `&IntoIter<PendingAttestation>` is not an iterator

Fwiw, if I alter the return type on `get_matching_source_attestations` to a simple `Vec`, `into_iter` (and everything else still) works -- so it seems like it's an `ssz_rs`-related aspect?

Generally speaking, is `into_iter` desired since it's more flexible than `iter`? E.g., yields `T`, `&T` or `&mut T` instead of only` &T`

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Sorry, idk if I follow -- do you mean binding to a variable instead of the `_` and then calling `drop()` on them?

ralexstokes: i mean like change the fn signature of `get_*_deltas` so that they only return one vec

no need to allocate the spare vec if we never use it -- ideally the rust compiler just even skips this but in case it does not (and leads to simpler code)

src/phase0/state_transition/epoch_processing.rs
ralexstokes: > Generally speaking, is into_iter desired since it's more flexible than iter? E.g., yields T, &T or &mut T instead of only &T

that is perhaps one way of looking at it -- `into_iter` yields an iterator by value, so that you get T instead of `iter` over &T or `iter_mut` over &mut T; i suppose there are situations where the compiler will insert borrows for you to match the call site if you use `into_iter`

my preference is to use "by value" over "by ref" where we can, but it is just an aesthetic thing and may have no bearing as the compiler may turn around and do crazy stuff w/ the code based on the context... the aesthetics here are to follow "sources and sinks" -- don't leave a thing lying around if you can just send the whole thing into a (e.g. trailing) function call

ralexstokes: and yeah i think i mislead w/ the `into_iter` call, just `iter` is fine here

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Oh, duh -- makes total sense

buchhlz2:
Lmk if these latest changes are good to go...I think I've captured everything relative to the last review/comment

PR #47, opened 2022-01-07
feat: add fn body to `process_effective_balance_updates`

buchhlz2:


src/phase0/state_transition/epoch_processing.rs
```diff
@@ -254,8 +254,25 @@ pub fn process_effective_balance_updates<
...
+    // Update effective balances with hysteresis
+    for i in 0..state.validators.len() {
+        let validator = &mut state.validators[i];
+        let balance = &state.balances[i];
+        let hysteresis_increment =
+            context.effective_balance_increment / context.hysteresis_quotient;
+        let downward_threshold = hysteresis_increment * context.hysteresis_downward_multiplier;
+        let upward_threshold = hysteresis_increment * context.hysteresis_upward_multiplier;
+        if balance + downward_threshold < validator.effective_balance
+            || validator.effective_balance + upward_threshold < *balance
```
buchhlz2, line 268: General rust question here within the `if` statement where `balance` is a `&`:
1. I didn't have to dereference `balance` in `balance + downward_threshold < validator.effective_balance`
2. However, I did have to dereference `balance` in `validator.effective_balance + upward_threshold < *balance`

Is this because comparison operators will compare expressions of only the same types? 

i.e., the statement in (1) somehow converts the LHS `balance + downward_threshold` to a normal `u64` and compares that to the RHS, also a `u64`. And in (2), if balance is not dereferenced, then the LHS is a `u64` and RHS is `&u64`, which can't be compared.

ralexstokes: you just need `PartialOrd` implemented to give meaning to operators like `<`

but unless you have something like `PartialOrd<&T> for T` then `rustc` won't know how to handle `a < b` where `a: &T` and `b: T`.

`.` has some special magic for auto-dereferencing and taken together these explain what you see.

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -254,8 +254,25 @@ pub fn process_effective_balance_updates<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<(), Error> {
+    // Update effective balances with hysteresis
+    for i in 0..state.validators.len() {
+        let validator = &mut state.validators[i];
+        let balance = &state.balances[i];
```
ralexstokes, line 262: balance is a `Gwei` which is `Copy` so you can just do an assign here:
```suggestion
        let balance = state.balances[i];
```

and that will simplify everything else :)

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -254,8 +254,25 @@ pub fn process_effective_balance_updates<
...
+        let hysteresis_increment =
+            context.effective_balance_increment / context.hysteresis_quotient;
+        let downward_threshold = hysteresis_increment * context.hysteresis_downward_multiplier;
+        let upward_threshold = hysteresis_increment * context.hysteresis_upward_multiplier;
```
ralexstokes, line 266: these can be hoisted outside the loop to iterate over all the vals

ralexstokes (commented):
a few suggestions but otherwise looks good!

src/phase0/state_transition/epoch_processing.rs
buchhlz2: Ahh that's right, ugh. Got it.

buchhlz2:
FYI -- idk if it's some issue with my clippy locally, but after pulling the latest changes from main into this branch, I'm getting an error from a different commit. It looks like github isn't complaining, though, but I can't build without an error being thrown:

--> src/phase0/state_transition/block_processing.rs:379:5
    |
379 | /     body.proposer_slashings
380 | |         .iter_mut()
    | |___________________^ cannot borrow as mutable
help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `ssz_rs::List<ProposerSlashing, MAX_PROPOSER_SLASHINGS>

ralexstokes:
> idk if it's some issue with my cargo/clippy locally, bu

had to update `ssz_rs` -- i would do `cargo update ssz_rs` to pull the latest

it is tracked by a git dep at the moment and sometimes cargo doesn't notice the change

PR #48, opened 2022-01-08
add `process_attester_slashing` impl

ralexstokes:


src/phase0/state_transition/block_processing.rs
```diff
@@ -109,9 +113,48 @@ pub fn process_attester_slashing<
...
+        }
+    }
+
+    if !slashed_any {
+        return Err(Error::InvalidOperation(InvalidOperation::AttesterSlashing(
+            InvalidAttesterSlashing::NoSlashings(indices),
+        )));
+    }
+
     Ok(())
```
ralexstokes, line 158: note to self: unify arms here

PR #56, opened 2022-01-10
feat: add `process_justification_and_finalization` impl

buchhlz2:
WIP placeholder for completing fn `process_justification_and_finalization`


buchhlz2:
fn impls to-be included:
- `process_justification_and_finalization`
    - `get_attesting_balance`
        - `get_unslashed_attesting_indices`(not included in this PR but noting it exists as part of https://github.com/ralexstokes/ethereum_consensus/pull/46)
    - `weigh_justification_and_finalization`

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -519,6 +539,120 @@ pub fn process_participation_record_updates<
...
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    total_active_balance: Gwei,
+    previous_epoch_target_balance: Gwei,
+    current_epoch_target_balance: Gwei,
+    context: &Context,
+) -> Result<(), Error> {
+    let previous_epoch = get_current_epoch(state, context);
+    let current_epoch = get_current_epoch(state, context);
+    let old_previous_justified_checkpoint = state.previous_justified_checkpoint.clone();
```
buchhlz2, line 604: There are a few `clones()` through this fn. I *think* these first three should be "allowable" because of the subsequent mutations to both `state.previous_justified_checkpoint` and `state.current_justified_checkpoint`, so borrowing wouldn't make sense here.

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -519,6 +539,120 @@ pub fn process_participation_record_updates<
...
+                .expect("block root exists for checkpoint"),
+        };
+        state.justification_bits.set(0, true);
+    }
+
+    // Process finalizations
+    let bits = &state.justification_bits;
+    // The 2nd/3rd/4th most recent epochs are justified, the 2nd using the 4th as source
+    if bits[1..4].all() && old_previous_justified_checkpoint.epoch + 3 == current_epoch {
+        state.finalized_checkpoint = old_previous_justified_checkpoint.clone();
```
buchhlz2, line 638: Here is where the other `clone`s are. I tried to at least reduce their usage by moving `old_previous_justified_checkpoint` instead of cloning within the second `if` statement below.

Technically, couldn't these first two `if`s be combined into one, thus, eliminating the initial `clone`? And the same would go for the 3rd & 4th `if`s

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -519,6 +539,120 @@ pub fn process_participation_record_updates<
...
+    let old_previous_justified_checkpoint = state.previous_justified_checkpoint.clone();
+    let old_current_justified_checkpoint = state.current_justified_checkpoint.clone();
+
+    // Process justifications
+    state.previous_justified_checkpoint = state.current_justified_checkpoint.clone();
+    for i in (1..(JUSTIFICATION_BITS_LENGTH - 1)).rev() {
+        let prev = state
+            .justification_bits
+            .get(i - 1)
+            .expect("at least one bit in justification_bits");
```
buchhlz2, line 613: A few instances of `expect` starting here -- calling out to double check if this is the desired way to handle it vs. some `Error` message.

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -5,9 +5,9 @@ use crate::phase0::state_transition::{
     get_block_root_at_slot, get_current_epoch, get_previous_epoch, get_randao_mix,
     get_total_active_balance, get_total_balance, get_validator_churn_limit, increase_balance,
     initiate_validator_exit, is_active_validator, is_eligible_for_activation,
-    is_eligible_for_activation_queue, Context, Error,
+    is_eligible_for_activation_queue, Checkpoint, Context, Error,
```
buchhlz2, line 8: I noticed `Copy` isn't derived for `Checkpoint` nor, mostly, any of the other types in this project. Just wondering what the design rationale is there (for my own curiosity/knowledge purposes).

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -519,6 +539,120 @@ pub fn process_participation_record_updates<
...
+            root: *get_block_root(state, current_epoch, context)
+                .expect("block root exists for checkpoint"),
+        };
+        state.justification_bits.set(0, true);
+    }
+
+    // Process finalizations
+    let bits = &state.justification_bits;
+    // The 2nd/3rd/4th most recent epochs are justified, the 2nd using the 4th as source
+    if bits[1..4].all() && old_previous_justified_checkpoint.epoch + 3 == current_epoch {
```
buchhlz2, line 637: With `JUSTIFICATION_BITS_LENGTH` being a constant, I was wondering if the hardcoded `bits[1..4]` is supposed to be used vs. `bits[1..JUSTIFICATION_BITS_LENGTH]`

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -179,8 +179,28 @@ pub fn process_justification_and_finalization<
         MAX_VALIDATORS_PER_COMMITTEE,
         PENDING_ATTESTATIONS_BOUND,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<(), Error> {
+    // Initial FFG checkpoint values have a `0x00` stub for `root`.
+    // Skip FFG updates in the first two epochs to avoid corner cases that might result in modifying this stub.
+    if get_current_epoch(state, context) > GENESIS_EPOCH + 1 {
```
buchhlz2, line 186: Noting the spec uses an early return:
```
if get_current_epoch(state) <= GENESIS_EPOCH + 1:
     return
```
...and I implemented the other side of this inequality as I wasn't sure if trying to do the python way is ideal here.

src/phase0/state_transition/epoch_processing.rs
ralexstokes: `Copy` should only be derived for types that are "lightweight" to copy

we could add `Copy` to `Checkpoint` but it is more uniform to just derive `Clone` -- we could go and make a separate pass to restrict types that derive Clone to deriving Copy but i don't know if it will gain a lot at the moment

we definitely want to use `Clone` when the type allocates, like if it includes a List or Vector SSZ type

src/phase0/state_transition/epoch_processing.rs
ralexstokes: i think it is easier to follow if we stick to the spec

i also prefer to keep the bodies of various blocks smaller rather than larger (which the spec does here)

src/phase0/state_transition/epoch_processing.rs
ralexstokes: looks good

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -179,8 +179,28 @@ pub fn process_justification_and_finalization<
...
+        let previous_attestations =
+            get_matching_target_attestations(state, get_previous_epoch(state, context), context);
+        let current_attestations =
+            get_matching_target_attestations(state, get_current_epoch(state, context), context);
+        let total_active_balance = get_total_active_balance(state, context)?;
+        let previous_target_balance =
+            get_attesting_balance(state, previous_attestations?, context)?;
```
ralexstokes, line 193: note: it is more idiomatic to use the `?` operator at the end of the variable's defn, not at a later use site

src/phase0/state_transition/epoch_processing.rs
ralexstokes: I think having an error for being out of bounds is preferable so we don't panic here

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -519,6 +539,120 @@ pub fn process_participation_record_updates<
...
+    let old_current_justified_checkpoint = state.current_justified_checkpoint.clone();
+
+    // Process justifications
+    state.previous_justified_checkpoint = state.current_justified_checkpoint.clone();
+    for i in (1..(JUSTIFICATION_BITS_LENGTH - 1)).rev() {
+        let prev = state
+            .justification_bits
+            .get(i - 1)
+            .expect("at least one bit in justification_bits");
+        state.justification_bits.set(i, prev);
```
ralexstokes, line 614: going to see if there isn't some kind of shift operation so we don't have to go bit-by-bit

src/phase0/state_transition/epoch_processing.rs
```diff
@@ -519,6 +539,120 @@ pub fn process_participation_record_updates<
...
+    }
+    state.justification_bits.set(0, false);
+    if previous_epoch_target_balance * 3 >= total_active_balance * 2 {
+        state.current_justified_checkpoint = Checkpoint {
+            epoch: previous_epoch,
+            root: *get_block_root(state, previous_epoch, context)
+                .expect("block root exists for checkpoint"),
```
ralexstokes, line 621: I think we can still do `?` here instead of `expect`

this is even more dangerous than the one above with the bits

src/phase0/state_transition/epoch_processing.rs
ralexstokes: ah good catch, this could use the constant, but it doesn't :)

there is a bit more symmetry in the code w/ digits here, rather than working in offsets from the constant

src/phase0/state_transition/epoch_processing.rs
ralexstokes: I think the compiler will complain unless you do it this way as you would potentially move the data in the first of each pair and then you would not have the data to move in the second of each pair

I don't think you can merge each set of pairs as these are very particular conditions as written and they don't hold if you start breaking up the logic

ralexstokes (approved):
overall looks great :)

sorry for the delay on review!

I have some minor stylistic things I want to change but the logic/impl looks good overall so I will merge now and then follow up with some minor changes later

src/phase0/state_transition/epoch_processing.rs
ralexstokes: https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#method.copy_within

buchhlz2:
All good! I'm sure it's been a busy past couple of months with the EIP, light client stuff, Devconnect, etc. Happy to help if you get back into things, though.

PR #57, opened 2022-01-19
WIP: add test cases for `process_block_header`

Cravtos:
There are a lot of commits in this PR because of the merge, but all changes I did are in 51d2640

Should I compare returned errors types to the expected ones? Also, not all returned errors are expected ones, for example in `test_invalid_parent_root` `Err(Shuffle)` is returned.


ralexstokes:
i'm glad you are proceeding w/ this test but can you cherry pick 51d2640 into its own branch and re-open a PR w/ that?

i left some comments on the commit

it will also be worth shoring up the testing infra broadly before charging ahead here much further -- i will prioritize getting #17 merged and then seeing how we want to run the `operations` tests

PR #63, opened 2022-02-01
feat: weak subjectivity impl

buchhlz2:
- Add fns `compute_weak_subjectivity_period` and `is_within_weak_subjectivity_period`
- Add helper fn `get_current_slot` -- used in `is_within_weak_subjectivity_period`
- Create types or consts for `Ether`, `ETH_TO_GWEI`, `SAFETY_DECAY`

### DIFF
--- a/src/phase0/weak_subjectivity.rs
+++ b/src/phase0/weak_subjectivity.rs
@@ -1,0 +1,134 @@
+/*
+use crate::phase0::beacon_state::BeaconState;
+use crate::phase0::state_transition::{
+    compute_epoch_at_slot, get_active_validator_indices, get_current_epoch,
+    get_total_active_balance, get_validator_churn_limit, Context, Error,
+};
+use crate::primitives::Slot;
+
+pub type Ether = u64;
+pub const ETH_TO_GWEI: u64 = 10u64.pow(9);
+pub const SAFETY_DECAY: u64 = 10;
+
+pub fn compute_weak_subjectivity_period<
+    const SLOTS_PER_HISTORICAL_ROOT: usize,
+    const HISTORICAL_ROOTS_LIMIT: usize,
+    const ETH1_DATA_VOTES_BOUND: usize,
+    const VALIDATOR_REGISTRY_LIMIT: usize,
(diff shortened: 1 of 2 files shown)


buchhlz2:
So, I started at least with what the [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/weak-subjectivity.md) has defined: custom types, consts, configs, and a couple of fns. I wasn't sure where to place everything within the overall context of the project -- such as altering `primitives`, `Config`, etc. -- so I've kept them all contained in a single `weak_subjectivity` module. 

For `is_within_weak_subjectivity_period`, it's supposed to take a `Store` as a param. I wasn't sure what to put here quite yet, so I used a `BeaconState` to represent it for now. Subsequently, the `get_current_slot` helper fn (which isn't defined in the spec) also takes `BeaconState` instead of `Store` and just returns its current slot.

src/phase0/weak_subjectivity.rs
```diff
@@ -0,0 +1,134 @@
...
+        HISTORICAL_ROOTS_LIMIT,
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    context: &Context,
+) -> Result<bool, Error> {
```
buchhlz2, line 91: There were some optional assertions that I plan on adding -- was going to do something like this but wanted to commit the basics before going any deeper:
```
if ws_state.latest_block_header.state_root != ws_checkpoint.root {
    Err(Error::...)
}
if compute_epoch_at_slot(ws_state.slot) != ws_checkpoint.epoch {
    Err(Error::...)
}
```

buchhlz2:
Also -- I just realized that _I think_ you meant to just use this project for the WS stuff but not open a PR like this lol. So lmk if that's correct and I'll delete these.

PR #66, opened 2022-02-12
Fix `index` check in `compute_shuffled_index`

Cravtos:
According to spec, `index` should be less than `index_count`. But `None` is returned if so.
```python3
def compute_shuffled_index(index: uint64, index_count: uint64, seed: Bytes32) -> uint64:
    assert index < index_count
    # ...
```


ralexstokes (approved):
nice catch!

PR #70, opened 2022-02-18
Fix `get_seed` fn

Cravtos:
Closes https://github.com/ralexstokes/ethereum_consensus/issues/67

The `epoch` variable was overridden with the epoch for getting randao mix.

According to the spec, this shouldn't happen:
```python3
def get_seed(state: BeaconState, epoch: Epoch, domain_type: DomainType) -> Bytes32:
    """
    Return the seed at ``epoch``.
    """
    mix = get_randao_mix(state, Epoch(epoch + EPOCHS_PER_HISTORICAL_VECTOR - MIN_SEED_LOOKAHEAD - 1))  # Avoid underflow
    return hash(domain_type + uint_to_bytes(epoch) + mix)
```

### DIFF
--- a/src/phase0/state_transition/mod.rs
+++ b/src/phase0/state_transition/mod.rs
@@ -847,9 +847,9 @@ pub fn get_seed<
     domain_type: DomainType,
     context: &Context,
 ) -> Bytes32 {
-    let epoch =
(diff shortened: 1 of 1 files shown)


ralexstokes:
another great catch!

PR #94, opened 2022-05-25
Remove the git submodule and add a justfile step to build integration tests

realbigsean:
The submodule makes using this as a dependency difficult because it takes a long time to download. And since it's only used in integration tests, it shouldn't be necessary when this is used as a dep.

~~Went with a simple solution here, to just clone the repo in a separate `build-integration-tests` `justfile` command. But it runs very slow. A better follow up solution might be porting over what we have in lighthouse because it is definitely faster: https://github.com/sigp/lighthouse/blob/stable/testing/ef_tests/Makefile  It doesn't map 1:1 with how it would work with `justfile` though based on when I tried it out.~~

I realized the tests here only use `general` vectors, so only downloading `general.tar.gz` from the spec test repo ends up being way faster.


ralexstokes (approved):
thank you!

lgtm :)

Issue #21, opened 2021-11-13
Implement the altair spec


ralexstokes:
The altair spec, e.g. here: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md

We will want to leverage as much of the phase 0 spec code as possible and this may require some care to make sure we can leverage multiple types etc. like an `apply_block` function that can take either spec's state, block types, or various functions that take a `Context` for each version of the spec

buchhlz2:
An overview of what's to be included:
- custom types
    - [ ]  ParticipationFlags
- constants
    - [ ]  TIMELY_SOURCE_FLAG_INDEX
    - [ ]  TIMELY_TARGET_FLAG_INDEX
    - [ ]  TIMELY_HEAD_FLAG_INDEX
    - [ ]  TIMELY_SOURCE_WEIGHT
    - [ ]  TIMELY_TARGET_WEIGHT
    - [ ]  TIMELY_HEAD_WEIGHT
    - [ ]  SYNC_REWARD_WEIGHT
    - [ ]  PROPOSER_WEIGHT
    - [ ]  WEIGHT_DENOMINATOR
    - [ ]  DOMAIN_SYNC_COMMITTEE
    - [ ]  DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF
    - [ ]  DOMAIN_CONTRIBUTION_AND_PROOF
    - [ ]  PARTICIPATION_FLAG_WEIGHTS
- preset
    - [ ]  INACTIVITY_PENALTY_QUOTIENT_ALTAIR
    - [ ]  MIN_SLASHING_PENALTY_QUOTIENT_ALTAIR
    - [ ]  PROPORTIONAL_SLASHING_MULTIPLIER_ALTAIR
    - [ ]  SYNC_COMMITTEE_SIZE
    - [ ]  EPOCHS_PER_SYNC_COMMITTEE_PERIOD
- config
    - [ ]  INACTIVITY_SCORE_BIAS
    - [ ]  INACTIVITY_SCORE_RECOVERY_RATE
- containers
    - [ ]  `BeaconBlockBody` → add `sync_aggregate`
    - [ ]  `BeaconState`
        - [ ]  add `inactivity_scores`, `current_sync_committee`, & `next_sync_committee`
        - [ ]  modify `previous_epoch_participation` & `current_epoch_participation`
    - [ ]  `SyncAggregate`
    - [ ]  `SyncCommittee`
- helper fns
[...]

PR #106, opened 2022-06-03
feat: implement `altair` spec

dtbuchholz:
This is a draft PR for https://github.com/ralexstokes/ethereum-consensus/issues/21 -- to implement the Altair spec. Overall, all of the fns have at least been implemented, except the fn mentioned below.

Namely, I'm having an issue with the `block_processing` fn `process_attestation` where I'm borrowing state as mutable (`epoch_participation`), and shortly thereafter, attempting to borrow it as immutable (`proposer_reward_numerator`) within a loop where both are required.

Aside from that, there are misc. instances where I've called out areas I was unsure about as well as a few `clone()`s to make things work. Search for comment leading with `@dev` for these callouts.

### DIFF
--- a/src/altair/block_processing.rs
+++ b/src/altair/block_processing.rs
@@ -225,7 +225,7 @@ pub fn process_deposit<
             state
                 .current_epoch_participation
                 .push(ParticipationFlags::default());
-            state.inactivity_scores.push(u64::default())
+            state.inactivity_scores.push(0)
         } else {
             return Err(invalid_operation_error(InvalidOperation::Deposit(
                 InvalidDeposit::InvalidSignature(deposit.data.signature.clone()),
--- a/src/altair/block_processing.rs
+++ b/src/altair/block_processing.rs
@@ -1,3 +1,7 @@
+// TODO remove once impl is added
+#![allow(dead_code)]
+#![allow(unused_mut)]
+#![allow(unused_variables)]
 use crate::altair as spec;
 
 use crate::crypto::fast_aggregate_verify;
@@ -14,14 +18,12 @@ use spec::{
     get_base_reward_per_increment, get_beacon_committee, get_beacon_proposer_index,
     get_block_root_at_slot, get_committee_count_per_slot, get_current_epoch, get_domain,
     get_indexed_attestation, get_previous_epoch, get_total_active_balance,
-    get_validator_from_deposit, has_flag, helpers::get_base_reward, increase_balance,
-    is_valid_indexed_attestation, process_block_header, process_eth1_data, process_operations,
-    process_randao, Attestation, BeaconBlock, BeaconState, Deposit, DepositMessage, SyncAggregate,
-    DEPOSIT_CONTRACT_TREE_DEPTH, PARTICIPATION_FLAG_WEIGHTS, PROPOSER_WEIGHT, SYNC_REWARD_WEIGHT,
-    WEIGHT_DENOMINATOR,
+    get_validator_from_deposit, has_flag, increase_balance, is_valid_indexed_attestation,
+    process_block_header, process_eth1_data, process_operations, process_randao, Attestation,
+    BeaconBlock, BeaconState, Deposit, DepositMessage, SyncAggregate,
 };
 use ssz_rs::prelude::*;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 use std::iter::zip;
 
 pub fn process_attestation<
@@ -124,54 +126,37 @@ pub fn process_attestation<
     let attesting_indices =
         get_attesting_indices(state, data, &attestation.aggregation_bits, context)?;
 
-    // @dev Create immutable reference to `{current_or_previous}_epoch_participation`
     let epoch_participation = if is_current {
-        &state.current_epoch_participation
-    } else {
-        &state.previous_epoch_participation
-    };
-
-    let mut proposer_reward_numerator = 0;
-    // @dev Use `participation_accumulator` to accumalate `index` and `ParticipationFlags` (via `add_flags`).
-    // Then, use this HashMap later in `set_epoch_participation` to achieve overwrite of the
-    // corresponding `{current_or_previous}_epoch_participation`'s flag.
-    // Replacement for the python code below, which would occur inside the loop's `if` statement:
-    // epoch_participation[index] = add_flag(epoch_participation[index], flag_index as u8);
-    let mut participation_accumulator = HashMap::<usize, ParticipationFlags>::default();
-    for index in attesting_indices {
-        for (flag_index, weight) in PARTICIPATION_FLAG_WEIGHTS.iter().enumerate() {
-            if participation_flag_indices.contains(&flag_index)
-                && !has_flag(epoch_participation[index], flag_index as u8)
-            {
-                participation_accumulator.insert(
-                    index,
-                    add_flag(epoch_participation[index], flag_index as u8),
-                );
-                // epoch_participation[index] = add_flag(epoch_participation[index], flag_index as u8);
-                // @dev explicit import of `get_base_reward` to disambiguate instead of `spec` import
-                // also, this is where the issue with mutable vs immutable borrow occurs
-
-                proposer_reward_numerator += get_base_reward(state, index, context)? * weight;
-            }
-        }
-    }
-
-    // @dev Create mutable reference to `{current_or_previous}_epoch_participation`
-    let set_epoch_participation = if is_current {
         &mut state.current_epoch_participation
     } else {
         &mut state.previous_epoch_participation
     };
-    // @dev Update flags in `set_epoch_participation` to corresponding values in `participation_accumulator`
-    for i in 0..set_epoch_participation.len() {
-        set_epoch_participation[i] = *participation_accumulator
-            .get(&i)
-            .expect("index in epoch participation should exist");
+
+    let mut proposer_reward_numerator = 0;
+    // @dev this is where I start having trouble
+    // the `get_base_reward` below borrows state as immutable, but `epoch_participation` above borrows as mutable
+    for index in attesting_indices {
+        for (flag_index, weight) in crate::altair::PARTICIPATION_FLAG_WEIGHTS.iter().enumerate() {
+            // if flag_index in participation_flag_indices and not has_flag(epoch_participation[index], flag_index):
+            if participation_flag_indices.contains(&flag_index)
+                && !has_flag(epoch_participation[index], flag_index as u8)
+            {
+                epoch_participation[index] = add_flag(epoch_participation[index], flag_index as u8);
+                // @dev explicit import of `get_base_reward` to disambiguate instead of `spec` import
+                // also, this is where the issue with mutable vs immutable borrow occurs
+                /*
+                proposer_reward_numerator +=
+                    crate::altair::helpers::get_base_reward(state, index, context)? * weight;
+                */
+            }
+        }
     }
 
     // Reward proposer
-    let proposer_reward_denominator =
-        (WEIGHT_DENOMINATOR - PROPOSER_WEIGHT) * WEIGHT_DENOMINATOR / PROPOSER_WEIGHT;
+    let proposer_reward_denominator = (crate::altair::WEIGHT_DENOMINATOR
+        - crate::altair::PROPOSER_WEIGHT)
+        * crate::altair::WEIGHT_DENOMINATOR
+        / crate::altair::PROPOSER_WEIGHT;
     let proposer_reward = proposer_reward_numerator / proposer_reward_denominator;
     increase_balance(
         state,
@@ -210,7 +195,7 @@ pub fn process_deposit<
         .map(|node| Node::from_bytes(node.as_ref().try_into().unwrap()))
         .collect::<Vec<_>>();
     let leaf = deposit.data.hash_tree_root()?;
-    let depth = DEPOSIT_CONTRACT_TREE_DEPTH + 1;
+    let depth = crate::altair::DEPOSIT_CONTRACT_TREE_DEPTH + 1;
     let index = state.eth1_deposit_index as usize;
     let root = &state.eth1_data.deposit_root;
     if !is_valid_merkle_branch(&leaf, branch.iter(), depth, index, root) {
@@ -337,24 +322,21 @@ pub fn process_sync_aggregate<
         get_total_active_balance(state, context)? / context.effective_balance_increment;
     let total_base_rewards =
         get_base_reward_per_increment(state, context)? * total_active_increments;
-    let max_participant_rewards =
-        total_base_rewards * SYNC_REWARD_WEIGHT / WEIGHT_DENOMINATOR / context.slots_per_epoch;
+    let max_participant_rewards = total_base_rewards * crate::altair::SYNC_REWARD_WEIGHT
+        / crate::altair::WEIGHT_DENOMINATOR
+        / context.slots_per_epoch;
     let participant_reward = max_participant_rewards / context.sync_committee_size as u64;
-    let proposer_reward =
-        participant_reward * PROPOSER_WEIGHT / (WEIGHT_DENOMINATOR - PROPOSER_WEIGHT);
+    let proposer_reward = participant_reward * crate::altair::PROPOSER_WEIGHT
+        / (crate::altair::WEIGHT_DENOMINATOR - crate::altair::PROPOSER_WEIGHT);
 
     // Apply participant and proposer rewards
-    let all_public_keys = state
-        .validators
-        .iter()
-        .enumerate()
-        .map(|(i, v)| (&v.public_key, i))
-        .collect::<HashMap<&BlsPublicKey, usize>>();
+    // @dev usage of clone here
+    let mut all_public_keys = state.validators.iter().map(|v| v.public_key.clone());
     let mut committee_indices: Vec<ValidatorIndex> = Vec::default();
     for public_key in state.current_sync_committee.public_keys.iter() {
         committee_indices.push(
-            *all_public_keys
-                .get(public_key)
+            all_public_keys
+                .position(|pk| pk == *public_key)
                 .expect("validator public_key should exist"),
         );
     }
--- a/src/altair/block_processing.rs
+++ b/src/altair/block_processing.rs
@@ -1,13 +1,30 @@
 // TODO remove once impl is added
 #![allow(dead_code)]
-
+#![allow(unused_mut)]
+#![allow(unused_variables)]
 use crate::altair as spec;
 
-use crate::state_transition::{Context, Result};
-use spec::{
-    process_block_header, process_eth1_data, process_operations, process_randao, Attestation,
-    BeaconBlock, BeaconState, Deposit, SyncAggregate,
+use crate::crypto::fast_aggregate_verify;
+use crate::domains::DomainType;
+use crate::primitives::{BlsPublicKey, ParticipationFlags, ValidatorIndex};
+use crate::signing::compute_signing_root;
+use crate::state_transition::{
+    invalid_operation_error, Context, InvalidAttestation, InvalidDeposit, InvalidOperation,
+    InvalidSyncAggregate, Result,
 };
+use spec::{
+    add_flag, compute_domain, compute_epoch_at_slot, decrease_balance,
+    get_attestation_participation_flag_indices, get_attesting_indices,
+    get_base_reward_per_increment, get_beacon_committee, get_beacon_proposer_index,
+    get_block_root_at_slot, get_committee_count_per_slot, get_current_epoch, get_domain,
+    get_indexed_attestation, get_previous_epoch, get_total_active_balance,
+    get_validator_from_deposit, has_flag, increase_balance, is_valid_indexed_attestation,
+    process_block_header, process_eth1_data, process_operations, process_randao, Attestation,
+    BeaconBlock, BeaconState, Deposit, DepositMessage, SyncAggregate,
+};
+use ssz_rs::prelude::*;
+use std::collections::HashSet;
+use std::iter::zip;
 
 pub fn process_attestation<
     const SLOTS_PER_HISTORICAL_ROOT: usize,
@@ -19,7 +36,7 @@ pub fn process_attestation<
     const MAX_VALIDATORS_PER_COMMITTEE: usize,
     const SYNC_COMMITTEE_SIZE: usize,
 >(
-    _state: &mut BeaconState<
+    state: &mut BeaconState<
         SLOTS_PER_HISTORICAL_ROOT,
         HISTORICAL_ROOTS_LIMIT,
         ETH1_DATA_VOTES_BOUND,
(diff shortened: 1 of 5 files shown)


src/altair/block_processing.rs
```diff
@@ -29,9 +46,123 @@ pub fn process_attestation<
...
+                /*
+                proposer_reward_numerator +=
+                    crate::altair::helpers::get_base_reward(state, index, context)? * weight;
+                */
+            }
+        }
+    }
+
+    // Reward proposer
+    let proposer_reward_denominator = (crate::altair::WEIGHT_DENOMINATOR
```
ralexstokes, line 156: cleaner to do this imo
```suggestion
    let proposer_reward_denominator = (WEIGHT_DENOMINATOR
```

and just pull all of these constants in from the `use spec::{..` up above

src/altair/block_processing.rs
```diff
@@ -55,9 +186,73 @@ pub fn process_deposit<
...
+    deposit: &mut Deposit,
+    context: &Context,
 ) -> Result<()> {
+    let branch = deposit
+        .proof
+        .iter()
+        .map(|node| Node::from_bytes(node.as_ref().try_into().unwrap()))
+        .collect::<Vec<_>>();
+    let leaf = deposit.data.hash_tree_root()?;
+    let depth = crate::altair::DEPOSIT_CONTRACT_TREE_DEPTH + 1;
```
ralexstokes, line 198:
```suggestion
    let depth = DEPOSIT_CONTRACT_TREE_DEPTH + 1;
```
ditto on imports of constants

src/altair/block_processing.rs
```diff
@@ -81,9 +276,86 @@ pub fn process_sync_aggregate<
...
+        )));
+    }
+
+    // Compute participant and proposer rewards
+    let total_active_increments =
+        get_total_active_balance(state, context)? / context.effective_balance_increment;
+    let total_base_rewards =
+        get_base_reward_per_increment(state, context)? * total_active_increments;
+    let max_participant_rewards = total_base_rewards * crate::altair::SYNC_REWARD_WEIGHT
+        / crate::altair::WEIGHT_DENOMINATOR
```
ralexstokes, line 326: constants here

src/altair/block_processing.rs
```diff
@@ -81,9 +276,86 @@ pub fn process_sync_aggregate<
...
+    // Apply participant and proposer rewards
+    // @dev usage of clone here
+    let mut all_public_keys = state.validators.iter().map(|v| v.public_key.clone());
+    let mut committee_indices: Vec<ValidatorIndex> = Vec::default();
+    for public_key in state.current_sync_committee.public_keys.iter() {
+        committee_indices.push(
+            all_public_keys
+                .position(|pk| pk == *public_key)
+                .expect("validator public_key should exist"),
+        );
```
ralexstokes, line 341: and then

```suggestion
        committee_indices.push(
            all_public_keys.get(public_key)
                .expect("validator public_key should exist"),
        );
```

src/altair/epoch_processing.rs
```diff
@@ -77,8 +138,27 @@ pub fn process_rewards_and_penalties<
...
+        return Ok(());
+    }
+
+    let mut deltas = Vec::new();
+    for flag_index in 0..crate::altair::PARTICIPATION_FLAG_WEIGHTS.len() {
+        let flag_index_delta = get_flag_index_deltas(state, flag_index, context)?;
+        deltas.push(flag_index_delta);
+    }
+    let mut inactivity_penalty_deltas = vec![get_inactivity_penalty_deltas(state, context)?];
+    deltas.append(&mut inactivity_penalty_deltas);
```
ralexstokes, line 155: I feel like we can skip the intermediate `Vec` allocation here

something like `deltas.append(get_inactivity_penalty_deltas(..)?);`

src/altair/epoch_processing.rs
```diff
@@ -102,8 +182,15 @@ pub fn process_participation_flag_updates<
...
-    _context: &Context,
 ) -> Result<()> {
+    let current_participation = mem::take(&mut state.current_epoch_participation);
+    state.previous_epoch_participation = current_participation;
+    // @dev intention is to set `rotate_participation` to a `List` of u8 values of length state's validators
+    // But, if `current_epoch_participation` is of length `VALIDATOR_REGISTRY_LIMIT` -- does this pose an issue?
+    let rotate_participation = vec![ParticipationFlags::default(); state.validators.len()];
+    state.current_epoch_participation = rotate_participation
+        .try_into()
+        .expect("should convert from Vec to List");
```
ralexstokes, line 193: can't we just do this ?
```suggestion
    state.current_epoch_participation = Default::default();
```

src/altair/epoch_processing.rs
```diff
@@ -102,8 +182,15 @@ pub fn process_participation_flag_updates<
         MAX_VALIDATORS_PER_COMMITTEE,
         SYNC_COMMITTEE_SIZE,
     >,
-    _context: &Context,
 ) -> Result<()> {
+    let current_participation = mem::take(&mut state.current_epoch_participation);
+    state.previous_epoch_participation = current_participation;
+    // @dev intention is to set `rotate_participation` to a `List` of u8 values of length state's validators
+    // But, if `current_epoch_participation` is of length `VALIDATOR_REGISTRY_LIMIT` -- does this pose an issue?
```
ralexstokes, line 189: the list limit is a bound which could theoretically be met but doesn't imply the specific instance is anywhere near that size

src/altair/epoch_processing.rs
ralexstokes, line 193: this is probably fine, and we could add a better API to the `List` type to do this kind of thing

src/altair/helpers.rs
```diff
@@ -43,9 +49,38 @@ pub fn get_next_sync_committee_indices<
...
+) -> Result<Vec<ValidatorIndex>> {
+    // Return the sync committee indices, with possible duplicates, for the next sync committee.
+    let epoch = get_current_epoch(state, context) + 1;
+    let max_random_byte = u8::MAX as u64;
+    let active_validator_indices = get_active_validator_indices(state, epoch);
+    let active_validator_count = active_validator_indices.len();
+    let seed = get_seed(state, epoch, DomainType::SyncCommittee, context);
+    let i = 0;
+    let mut sync_committee_indices = Vec::<ValidatorIndex>::new();
+    // @dev need to validate this is correct for `hash`
```
ralexstokes, line 62: seems ok

src/altair/helpers.rs
```diff
@@ -68,9 +103,36 @@ pub fn get_next_sync_committee<
...
+        .filter_map(|(i, v)| {
+            if indices.contains(&i) {
+                Some(&v.public_key)
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
+    // @dev not sure if the impl of `eth_aggregate_pubkeys` is needed or correct
+    // also, not sure if `expect` is desired here
```
ralexstokes, line 123: why not call `eth_aggregate_pubkeys`?

also I would use `?` operator rather than `expect` here

src/state_transition/error.rs
```diff
@@ -212,7 +212,10 @@ pub enum InvalidVoluntaryExit {
 
 #[derive(Debug, Error)]
 #[error("TODO: fill out")]
```
ralexstokes, line 214:
```suggestion
```

src/crypto.rs
```diff
@@ -86,6 +86,19 @@ pub fn fast_aggregate_verify(pks: &[&PublicKey], msg: &[u8], signature: &Signatu
...
 
+// @dev not sure if this is the correct impl
+pub fn eth_aggregate_pubkeys(pks: &[&PublicKey]) -> Result<PublicKey, Error> {
+    // Return the aggregate public key for the public keys in `pks`
+    if pks.is_empty() {
+        return Err(Error::EmptyInput);
+    }
+    let v: Vec<&blst_core::PublicKey> = pks.iter().map(|pk| &pk.0).collect();
+
+    blst_core::AggregatePublicKey::aggregate(&v, true)
```
ralexstokes, line 97: can double-check the apis but this looks about right

ralexstokes (commented):
excellent start!

make a shallow pass and left some comments

one general point: I would import all of the various constants directly from the `spec` module and then you can drop the `crate::altair` imports inline

if these values are used, but changed in a future fork, it will be transparent to this calling code to change them if they are hidden behind the `spec` module

src/altair/block_processing.rs
dtbuchholz: Ahh, okay, yeah all of these were based on this comment, which I misinterpreted:
https://github.com/ralexstokes/ethereum-consensus/blob/65fc02162a59d70106f73f6b6984167161cb1c06/src/altair/helpers.rs#L229

In the example above, the const is _passed to a fn_ -- this is the only time where direct imports should be used for code gen simplification purposes. Otherwise, like you said, just do a standard top-level import.

src/altair/helpers.rs
dtbuchholz: Oops -- should have clarified. In `crypto`, I've added the impl for `eth_aggregate_pubkeys` (which I'll rename to `eth_aggregate_public_keys` for consistency purposes in this library). I'm not sure if _this_ is the correct impl. Looks like it is (potentially) correct via your comment https://github.com/ralexstokes/ethereum-consensus/pull/106/files#r888586787

dtbuchholz: Re: `?` operator -- I get an error if I try to do this: `'?' couldn't convert the error to 'state_transition::error::Error'`

```
the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
the following implementations were found:
  <state_transition::error::Error as std::convert::From<ssz_rs::MerkleizationError>>
  <state_transition::error::Error as std::convert::From<ssz_rs::SimpleSerializeError>>
  <state_transition::error::Error as std::convert::From<std::boxed::Box<state_transition::error::InvalidBlock>>>
required because of the requirements on the impl of `std::ops::FromResidual<std::result::Result<std::convert::Infallible, crypto::Error>>` for `std::result::Result<altair::sync::SyncCommittee<SYNC_COMMITTEE_SIZE>, state_transition::error::Error>`rustc[E0277](https://doc.rust-lang.org/error-index.html#E0277)
helpers.rs(107, 6): expected `state_transition::error::Error` because of this
```

The only "new" thing `eth_aggregate_pubkeys` uses is `AggregatePublicKey::aggregate`, which has the same `BLSTError` as the other fns in `crypto`, so I don't think that's the issue. It also returns a Result with `PublicKey`, which I don't think any others in `crypto` return. Is that potentially the issue? E.g., `state_transition::error` needs something like `InvalidPublicKey(BlsPublicKey)`?

Otherwise, it looks like it's coming from `SyncCommittee<SYNC_COMMITTEE_SIZE>, state_transition::error::Error`?

src/altair/block_processing.rs
dtbuchholz: Although, another fn uses the direct imports mentioned here:
https://github.com/ralexstokes/ethereum-consensus/blob/65fc02162a59d70106f73f6b6984167161cb1c06/src/altair/helpers.rs#L297

Is there a rule of thumb between using direct imports (like you've done) vs. top-level?

src/altair/helpers.rs
```diff
@@ -68,9 +103,36 @@ pub fn get_next_sync_committee<
...
+    let aggregate_public_key = eth_aggregate_pubkeys(&public_keys)
+        .expect("validator public_keys should be aggregated into aggregate_public_key");
+
+    // @dev ran into trouble here so just cloned() the validator's public_key
+    // couldn't figure out how to properly coerce/convert `public_keys` Vec to ssz_rs::Vector
+    let pks_vector =
+        Vector::<_, SYNC_COMMITTEE_SIZE>::from_iter(public_keys.iter().map(|&pk| pk.clone()));
```
dtbuchholz, line 130: Note: fixed with `public_keys.iter().map(|&pk| *pk`

dtbuchholz: Jk nvm lol 
```
cannot move out of `*pk` which is behind a shared reference
move occurs because `*pk` has type `crypto::PublicKey`, which does not implement the `Copy` trait
```

src/altair/block_processing.rs
```diff
@@ -29,9 +46,123 @@ pub fn process_attestation<
...
+        get_attesting_indices(state, data, &attestation.aggregation_bits, context)?;
+
+    let epoch_participation = if is_current {
+        &mut state.current_epoch_participation
+    } else {
+        &mut state.previous_epoch_participation
+    };
+
+    let mut proposer_reward_numerator = 0;
+    // @dev this is where I start having trouble
```
dtbuchholz, line 136: I at least got something to work using a `HashMap`, similar to what was recommended in `process_sync_aggregate`. Can push once these other cleanups are confirmed...but here's what I've done:
https://gist.github.com/dtbuchholz/9ca416cda0d4b368d873bedee8c16eef

dtbuchholz:
(nre: accidentally pushed some stuff to my repo; ignore)

src/altair/block_processing.rs
ralexstokes: tbh not sure why I did this other than taking a confusing shortcut...

I can patch this up, no reason these shouldn't flow forward to each fork

src/altair/helpers.rs
ralexstokes: yeah I'm a little surprised this is not already there but we want to do something like this:

https://github.com/ralexstokes/ethereum-consensus/blob/main/src/state_transition/error.rs#L13-L14

so there is another enum variant `Crypto(#[from] CrytpoError)` if you do something like `use crate::crypto::Error as CryptoError;`

src/altair/helpers.rs
```diff
@@ -68,9 +103,36 @@ pub fn get_next_sync_committee<
...
+        .iter()
+        .enumerate()
+        .filter_map(|(i, v)| {
+            if indices.contains(&i) {
+                Some(&v.public_key)
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
```
ralexstokes, line 121: try this:

```suggestion
                Some(v.public_key.clone())
            } else {
                None
            }
        })
        .collect::<Vector<_, SYNC_COMMITTEE_SIZE>>();
```

ralexstokes: you can skip one intermediate allocated `Vec` this way

and it is ok to clone the public keys here but there will just be different sets

you can imagine some kind of optimized implementation that only tracks indices of pubkeys but simpler to just copy them over for now

ralexstokes: oh also to be clear you can drop the line below that tries to build `public_keys`

src/altair/block_processing.rs
```diff
@@ -29,9 +46,123 @@ pub fn process_attestation<
...
+        )));
+    }
+
+    // Participation flag indices
+    let inclusion_delay = state.slot - data.slot;
+    let participation_flag_indices =
+        get_attestation_participation_flag_indices(state, data, inclusion_delay, context)?;
+
+    // Verify signature
+    let _ = is_valid_indexed_attestation(
```
ralexstokes, line 118:
```suggestion
    is_valid_indexed_attestation(
```

ralexstokes: this is kind of weird and really should prolly return a boolean given how it reads but not super important imo

src/altair/block_processing.rs
```diff
@@ -29,9 +46,123 @@ pub fn process_attestation<
...
+        state,
+        &mut get_indexed_attestation(state, attestation, context)?,
+        context,
+    )?;
+
+    // Update epoch participation flags
+    // @dev deviate from the order of the spec to avoid immutable borrow after mutable borrow
```
ralexstokes, line 125: I'm confused where the conflicting borrow is

src/altair/block_processing.rs
ralexstokes: yeah this seems like it would work!

I think we may be able to coerce the borrow checker some and get something a bit simpler, let me try something

ralexstokes: so this is a bit annoying but I would just do something like this:

https://gist.github.com/rust-play/149e04c2a657723f4d3fb477208881e3

the borrow checker doesn't like that we have a mutable borrow into `state` across the span of the loop where we also pass a second mutable borrow of the entire thing

if we had mutable borrows of disjoint fields of `state` this would be fine but I don't think it is worth rewriting `get_base_reward` at this time

src/altair/helpers.rs
dtbuchholz: ahh that was easy lol

src/altair/block_processing.rs
dtbuchholz: ahh yeah, much simpler

dtbuchholz:
Ok, so I believe I've addressed everything except the `let _ = is_valid_indexed_attestation` to remove the weird allocation in `block_processing`...which I didn't see your comment until after the push. And since it's not a huge deal, I've left it as is. But nbd if you want me to create a quick cleanup.

Also, it's been a minute since I merged something with a lotta back/forth changes. Regarding squashing my "messy" commits into one -- would you like me to do that here? 

Last thing -- sorry for the hold up; the past 1-2 weeks was our seasonal planning/strategy, so it took an unusual amount of time up in addition to some nft.nyc aspects leading up to it this week. Aka I should be able to move a lil quicker moving forward.

ralexstokes:
sweet I'll review later today!

I do prefer fewer commits and I'd prolly rebase all of these into one commit but I wouldn't block merging on this... but esp. stuff like the revert commits, we can just edit the history so they don't even exist in the first place

not a problem! I don't expect any speed here and I am grateful for any contributions at all :) thank you!!

src/crypto.rs
```diff
@@ -86,6 +86,20 @@ pub fn fast_aggregate_verify(pks: &[&PublicKey], msg: &[u8], signature: &Signatu
     res == BLST_ERROR::BLST_SUCCESS
 }
 
+pub fn eth_aggregate_public_keys<const N: usize>(
+    pks: &Vector<PublicKey, N>,
```
ralexstokes, line 90:
```suggestion
    pks: &[PublicKey],
```

and then we can drop the const generic param here

I think here we are fine to just take a slice of keys, it may mean callers get a bit more complex but that's ok imo

src/altair/helpers.rs
```diff
@@ -292,6 +450,7 @@ pub fn slash_validator<
 
     let whistleblower_reward =
         state.validators[slashed_index].effective_balance / context.whistleblower_reward_quotient;
+    // NOTE: direct imports to simplify forward code gen of these constants
```
ralexstokes, line 453:
```suggestion
```

p sure we want this

src/altair/helpers.rs
```diff
@@ -144,11 +204,47 @@ pub fn get_unslashed_participating_indices<
...
+                None
+            }
+        })
+        .collect::<Vec<_>>();
+    let unslashed_particpating_indices = state
+        .validators
+        .iter()
+        .enumerate()
+        .filter_map(|(i, v)| if !v.slashed { Some(i) } else { None })
+        .filter(|i| participating_indices.contains(i))
```
ralexstokes, line 245: looks like this is equivalent
```suggestion
        .filter(|(i, v)| !v.slashed && participating_indices.contains(i))
```

src/altair/helpers.rs
```diff
@@ -144,11 +204,47 @@ pub fn get_unslashed_participating_indices<
...
+            requested: epoch,
+            previous: previous_epoch,
+            current: current_epoch,
+        });
+    }
+
+    let active_validator_indices = get_active_validator_indices(state, epoch);
+    let participating_indices = active_validator_indices
+        .iter()
+        .filter_map(|&i| {
```
ralexstokes, line 226:
```suggestion
        .filter(|&i| {
```

and then simplify the predicate body

src/altair/helpers.rs
```diff
@@ -69,9 +103,28 @@ pub fn get_next_sync_committee<
...
+        .enumerate()
+        .filter_map(|(i, v)| {
+            if indices.contains(&i) {
+                Some(v.public_key.clone())
+            } else {
+                None
+            }
+        })
+        .collect::<Vector<_, SYNC_COMMITTEE_SIZE>>();
+    let aggregate_public_key = eth_aggregate_public_keys(&public_keys)?;
```
ralexstokes, line 122: if we change `eth_aggregate_public_keys` like I suggested below then I'm not sure we have `Deref` implemented for `Vector<T, _>` to `&[T]` but it would be useful imo

src/altair/helpers.rs
```diff
@@ -44,9 +50,37 @@ pub fn get_next_sync_committee_indices<
...
+    context: &Context,
+) -> Result<Vec<ValidatorIndex>> {
+    // Return the sync committee indices, with possible duplicates, for the next sync committee.
+    let epoch = get_current_epoch(state, context) + 1;
+    let max_random_byte = u8::MAX as u64;
+    let active_validator_indices = get_active_validator_indices(state, epoch);
+    let active_validator_count = active_validator_indices.len();
+    let seed = get_seed(state, epoch, DomainType::SyncCommittee, context);
+    let i = 0;
+    let mut sync_committee_indices = Vec::<ValidatorIndex>::new();
```
ralexstokes, line 62: we could do `Vec::with_capacity` as we know the length of this thing ahead of time

src/altair/helpers.rs
```diff
@@ -69,9 +103,28 @@ pub fn get_next_sync_committee<
...
+    context: &Context,
+) -> Result<SyncCommittee<SYNC_COMMITTEE_SIZE>> {
+    // Return the next sync committee, with possible pubkey duplicates.
+    let indices = get_next_sync_committee_indices(state, context)?;
+    let public_keys = state
+        .validators
+        .iter()
+        .enumerate()
+        .filter_map(|(i, v)| {
+            if indices.contains(&i) {
```
ralexstokes, line 115: also, if we are only ever using the `indices` as a set to check membership (I believe we are) then `get_next_sync_committee_indices` should collect into a `HashSet`

src/altair/epoch_processing.rs
```diff
@@ -127,8 +212,14 @@ pub fn process_sync_committee_updates<
         MAX_VALIDATORS_PER_COMMITTEE,
         SYNC_COMMITTEE_SIZE,
     >,
-    _context: &Context,
+    context: &Context,
 ) -> Result<()> {
+    let next_sync_committee = mem::take(&mut state.next_sync_committee);
```
ralexstokes, line 217: so what happens if we are not aligned w/ the sync committee boundary?

`mem::take` here would just clear out the `next_sync_committee` which we don't want

hint: move this line inside the fn body

src/altair/block_processing.rs
```diff
@@ -55,9 +175,73 @@ pub fn process_deposit<
...
+                depth,
+                index,
+                root: *root,
+            },
+        )));
+    }
+
+    // NOTE: deviate from the order of the spec to avoid mutations
+    // that would need to be rolled back upon failure
+    let public_key = deposit.data.public_key.clone();
```
ralexstokes, line 204:
```suggestion
    let public_key = &deposit.data.public_key;
```

it looks like this is fine and saves a clone :)

src/altair/block_processing.rs
```diff
@@ -55,9 +175,73 @@ pub fn process_deposit<
...
+    let amount = deposit.data.amount;
+    let validator_public_keys: HashSet<&BlsPublicKey> =
+        HashSet::from_iter(state.validators.iter().map(|v| &v.public_key));
+    if !validator_public_keys.contains(&public_key) {
```
ralexstokes, line 208: if so can drop the ref here
```suggestion
    if !validator_public_keys.contains(public_key) {
```

src/altair/block_processing.rs
```diff
@@ -55,9 +175,73 @@ pub fn process_deposit<
...
+                .validators
+                .push(get_validator_from_deposit(deposit, context));
+            state.balances.push(amount);
+            state
+                .previous_epoch_participation
+                .push(ParticipationFlags::default());
+            state
+                .current_epoch_participation
+                .push(ParticipationFlags::default());
+            state.inactivity_scores.push(u64::default())
```
ralexstokes, line 228: I have a weak preference for this 
```suggestion
            state.inactivity_scores.push(0)
```

but tbh the compiler probably compiles both to the same code

ralexstokes (commented):
excellent! this is a big one haha

made another pass and left some light feedback

ralexstokes:
closing in lieu of #116

Issue #113, opened 2022-06-22
Implement the execution logic for the spec tests


ralexstokes:
The skeleton for the spec tests was added in #112.

There are `rustc` tests that call each of the test cases in the locally provided `consensus-spec-tests`.

Right now, most of the different test runners/handlers are merely stubbed out and we need to implement the rest of the functionality.

the outstanding handlers:

- [x] [genesis](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/genesis)
- [x] [forks](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/forks)
- [x] [transition](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/transition)
- [x] [rewards](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/rewards)
- [x] [operations](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/operations)
- [x] [bls](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/bls)
- [x] [ssz_static](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/ssz_static)
- [x] [epoch_processing](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/epoch_processing)
- [x] [finality](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/finality)
- [x] [random](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/random)
- [x] [sanity](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/sanity)
- [x] [shuffling](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/shuffling)

please post below if you want to tackle one of them -- I think something roughly like 1 PR per handler type makes a lot of sense.

ralexstokes:
#5 is relevant

ralexstokes:
I'm going to work on the `epoch_processing` tests next to get a sense for any other abstractions I think we may want, along w/ exercising some of the core of the spec code


@pkopparla if you wanted to grab another handler the `ssz_static` ones should be pretty straightforward and I don't think would need to change much if other parts of the handlers do change

ralexstokes:
closed in #138

PR #114, opened 2022-06-24
Phase 0 bls tests

pkopparla:
Addresses item 1 of #113 
BLS tests with the new test runner format.

### DIFF
--- a/tests/spec_test_runners/bls/mod.rs
+++ b/tests/spec_test_runners/bls/mod.rs
@@ -1,47 +1,174 @@
 use crate::test_utils::{load_yaml, TestCase};
 use ethereum_consensus::crypto::{
-    // aggregate, aggregate_verify, fast_aggregate_verify, PublicKey,
-    SecretKey,
-    Signature,
+    aggregate, aggregate_verify, fast_aggregate_verify, PublicKey, SecretKey, Signature,
 };
-// use ethereum_consensus::primitives::Bytes32;
+use ethereum_consensus::primitives::Bytes32;
 use ethereum_consensus::serde as eth_serde;
 use serde::Deserialize;
 use serde_with::{serde_as, DefaultOnError};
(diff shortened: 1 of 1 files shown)


tests/spec_test_runners/bls/mod.rs
```diff
@@ -1,47 +1,174 @@
...
-pub struct AggregateHandler;
+#[derive(Debug, Deserialize)]
+struct AggregateTestCase {
+    input: Vec<Signature>,
+    output: Option<Signature>,
+}
+
+#[derive(Debug)]
+pub struct AggregateHandler {
+    test_case: AggregateTestCase,
```
ralexstokes, line 18: I'm wondering if we need the extra indirection for `test_case` here

instead we could just inline the `input` and `output` members into the `Handler` types

I will go look at some of the other test types to get a clearer idea, otherwise it just bloats all of the code unnecessarily

tests/spec_test_runners/bls/mod.rs
```diff
@@ -1,47 +1,174 @@
...
+}
 
 impl AggregateHandler {
     pub fn from(test_case_path: &str) -> Self {
-        Self
+        let path = test_case_path.to_string() + "/data.yaml";
+        let test_case: AggregateTestCase = load_yaml(&path);
+        Self { test_case }
```
ralexstokes, line 25: we may also be able to make a default impl here to skip having to define all of this code for each thing

ralexstokes (approved):
looks great!

I do have some high-level questions but they will only be answered by writing some other handler types to better understand the constraints

I'll go ahead and merge this now and we can refactor later if needed

evm626:
I agree with your comments here. Which category of tests next?

pkopparla:
I agree with your comments here. Which category of tests next?

PR #139, opened 2022-07-28
put libp2p-core behind a feature

michaelsproul:
In Lighthouse we're adding `mev-rs` as a dependency, and bringing in `ethereum-consensus` is currently adding a duplicate version of `libp2p-core` that's different from upstream.

It seems that `mev-rs` doesn't need `PeerId`, so I've turned this feature _off_ by default. If you'd like a feature name other than `peer-id` let me know (maybe `libp2p`?).


ralexstokes:
cool!

this is a bit of a placeholder and I can say that pulling in this one type pulled in a *bunch* of extra deps, added to build times etc which I strongly don't want so I wanted to eventually replace w/ a custom defn -- and it is not even clear this repo needs this in the long run...


I'll merge this for now and make a note to clean this up for myself later,

thanks for the PR!

PR #159, opened 2022-09-09
upgrade enr, libp2p-core dependecies. Feature gate networking

divagant-martian:
Updates the enr and libp2p-core dependecies.
Adds a features `"networking"`, enabled by default for compatibility that gates the networking module. This allows users who don't need it to prevent bringin in the enr and libp2p-core dependencies at all


ralexstokes:
Closing this PR as I think we found another workaround; please feel free to re-open if needed :)

PR #161, opened 2022-09-15
+ `serde::as_string` to `Coordinate.slot`

jmcook1186:
adds `serde::as_string` to the slot field in the Coordinate struct to enable deserializing from json.


ralexstokes (approved):
nice!

PR #165, opened 2022-10-25
impl display for PeerId

jmcook1186:
implements `fmt::Display` on `PeerId`

### DIFF
--- a/src/networking.rs
+++ b/src/networking.rs
@@ -1,10 +1,13 @@
 use crate::primitives::Epoch;
 use crate::ssz::ByteList;
 use enr;
+pub use multiaddr::Multiaddr;
+use multihash::{Code, Error, Multihash};
 use ssz_rs::prelude::Bitvector;
 use std::fmt;
 use std::time::Duration;
-
+use std::{convert::TryFrom, str::FromStr};
+use thiserror::Error;
 pub const ATTESTATION_SUBNET_COUNT: usize = 64;
 pub const GOSSIP_MAX_SIZE: usize = 2usize.pow(20);
 pub const MAX_REQUEST_BLOCKS: usize = 2usize.pow(10);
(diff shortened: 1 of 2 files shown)


src/networking.rs
```diff
@@ -23,6 +24,12 @@ const MAX_PEER_ID_BYTE_COUNT: usize = 512;
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct PeerId(#[serde(with = "crate::serde::as_b58")] ByteList<MAX_PEER_ID_BYTE_COUNT>);
 
+impl fmt::Display for PeerId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
```
ralexstokes, line 29: as written, this will just call `Display::fmt` on the inner `ByteList` which I think will just dump the hex-encoded byte list

but instead we want the "canonical" base58 encoding so will want to do something like this

```suggestion
#[cfg(feature = "serde")]
impl fmt::Display for PeerId {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "{}", bs58::encode(self.0))
```

ralexstokes (commented):
nice! one minor fix to encode correctly

src/networking.rs
ralexstokes: hmm actually this also is wrapped in a "multi hash"

you can see the implementation here:

https://github.com/libp2p/go-libp2p/blob/v0.23.2/core/peer/peer.go#L179

I'd suggest looking for a `multihash` impl or referring to a rust libp2p impl to see how they implement this and just copy that code

jmcook1186:
updated this pr to reimplement `PeerId` using code from `libp2p`. The minimal version we discussed on tg broke serde so I added the explicit Ser/De. Added one test to check the output with the default formatter. All tests pass, but maybe there's a more elegant way to get around the ser/deser issue.

src/networking.rs
```diff
@@ -14,14 +17,120 @@ pub const RESP_TIMEOUT: Duration = Duration::from_secs(10);
...
 
-pub use multiaddr::Multiaddr;
+/// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
+/// automatically used as the peer id using an identity multihash.
+const MAX_INLINE_KEY_LENGTH: usize = 42;
 
-// TODO see how much this can be tightened
-const MAX_PEER_ID_BYTE_COUNT: usize = 512;
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
```
ralexstokes, line 25: let's move these so they are grouped w/ the rest of the imports

src/networking.rs
```diff
@@ -14,14 +17,120 @@ pub const RESP_TIMEOUT: Duration = Duration::from_secs(10);
 pub const ATTESTATION_PROPAGATION_SLOT_RANGE: usize = 32;
 pub const MAXIMUM_GOSSIP_CLOCK_DISPARITY: Duration = Duration::from_millis(500);
 
-pub use multiaddr::Multiaddr;
+/// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
```
ralexstokes, line 20: let's add a comment here that this was cribbed from `rust-libp2p` and that we should revisit the implementation

ralexstokes (commented):
excellent! I am not certain we will keep all of this but it unblocks other stuff right now so I say we move ahead...

I left a few comments on things to change but then I'll merge

src/networking.rs
jmcook1186: done!

src/networking.rs
jmcook1186: done!

PR #168, opened 2023-01-13
Add Capella types and presets

jimmygchen:
Add Capella types and presets from latest consensus specs
 https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#preset

PR #170, opened 2023-01-16
Add Deneb (EIP-4844) types and presets

jimmygchen:


src/kzg.rs
```diff
@@ -0,0 +1,144 @@
...
+            expected: KZG_COMMITMENT_BYTES_LEN,
+        })?;
+        Ok(Self(inner))
+    }
+}
+
+#[derive(Default, Debug, Clone, SimpleSerialize, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(into = "String", try_from = "String"))]
+pub struct KzgProof(ByteVector<KZG_PROOF_BYTES_LEN>);
```
ralexstokes, line 89: I think for now we can just get away with `pub type KzgProof = ByteVector<KZG_PROOF_BYTES_LEN>;`

and drop the rest of the code defining/impl'ing `KzgProof`

src/kzg.rs
```diff
@@ -0,0 +1,144 @@
...
+    #[error(
+        "invalid length of encoding: expected {expected} bytes but only provided {provided} bytes"
+    )]
+    EncodingError { provided: usize, expected: usize },
+}
+
+#[derive(Default, Debug, Clone, SimpleSerialize, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(into = "String", try_from = "String"))]
+pub struct KzgCommitment(ByteVector<KZG_COMMITMENT_BYTES_LEN>);
```
ralexstokes, line 29: I think for now we can just get away with `pub type KzgCommitment = ByteVector<KZG_COMMITmENT_BYTES_LEN>;`

and drop the rest of the code defining/impl'ing `KzgCommitment`

src/kzg.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Debug, Error)]
+pub enum Error {
+    #[cfg(feature = "serde")]
+    #[error("error deserializing hex-encoded input: {0}")]
+    Hex(#[from] HexError),
+    #[error(
+        "invalid length of encoding: expected {expected} bytes but only provided {provided} bytes"
+    )]
+    EncodingError { provided: usize, expected: usize },
+}
```
ralexstokes, line 24: do we need to keep this custom `Error` if we just use the simple type aliases below?

ralexstokes (commented):
I'm going to focus on #168 first but I skimmed this

directionally, it looks good!

src/kzg.rs
jimmygchen: perfect, thanks!

src/kzg.rs
jimmygchen: 👍

src/kzg.rs
jimmygchen: Removed :) the less code the better

src/kzg.rs
```diff
@@ -0,0 +1,10 @@
+use crate::primitives::Bytes32;
+use crate::ssz::ByteVector;
+use ssz_rs::prelude::*;
+
+const KZG_COMMITMENT_BYTES_LEN: usize = 48;
+const KZG_PROOF_BYTES_LEN: usize = 48;
```
ralexstokes, line 6: a bit random but may be worth making these `pub`, in the event some user of the library needed them for some reason...

jimmygchen: Done. Make sense 👍

jimmygchen:
Will revisit this PR once the spec changes for decoupling blobs (https://hackmd.io/cmYisgxkRuGe9NjX4gr97A?view) is closer to ready.

jimmygchen:
Hi @ralexstokes 
I have made another update:
- replace all EIP-4844 references with Deneb
- implement latest changes from blobs decoupling (https://github.com/ethereum/consensus-specs/pull/3244)
- add blinded blob types

src/configs/sepolia.rs
```diff
@@ -20,8 +20,8 @@ pub const BELLATRIX_FORK_VERSION: Version = [144, 0, 0, 113];
 pub const BELLATRIX_FORK_EPOCH: Epoch = 100;
 pub const CAPELLA_FORK_VERSION: Version = [3, 0, 16, 32];
 pub const CAPELLA_FORK_EPOCH: Epoch = FAR_FUTURE_EPOCH;
-pub const EIP4844_FORK_VERSION: Version = [4, 0, 16, 32];
-pub const EIP4844_FORK_EPOCH: Epoch = FAR_FUTURE_EPOCH;
+pub const DENEB_FORK_VERSION: Version = [4, 0, 16, 32];
+pub const DENEB_FORK_EPOCH: Epoch = FAR_FUTURE_EPOCH;
```
ralexstokes, line 24: the `fork_version` looks incorrect (seems like the goerli values)

ill go fix in another PR and also set the sepolia fork epoch, since we know it now

ralexstokes (commented):
nice updates!

src/configs/sepolia.rs
jimmygchen: I've created a PR to update the sepolia values for Capella
https://github.com/ralexstokes/ethereum-consensus/pull/194

Doesn't look like the Deneb config are out yet, so I'll leave them as it is.

src/deneb/beacon_block.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub sync_aggregate: SyncAggregate<SYNC_COMMITTEE_SIZE>,
+    pub execution_payload: ExecutionPayload<
+        BYTES_PER_LOGS_BLOOM,
+        MAX_EXTRA_DATA_BYTES,
+        MAX_BYTES_PER_TRANSACTION,
+        MAX_TRANSACTIONS_PER_PAYLOAD,
+        MAX_WITHDRAWALS_PER_PAYLOAD,
+    >,
+    pub bls_to_execution_changes: List<SignedBlsToExecutionChange, MAX_BLS_TO_EXECUTION_CHANGES>,
+    pub blob_kzg_commitments: List<KzgCommitment, MAX_BLOBS_PER_BLOCK>,
```
jimmygchen, line 47: May have to update `MAX_BLOBS_PER_BLOCK` to `MAX_BLOB_COMMITMENTS_PER_BLOCK`. 
https://github.com/ethereum/consensus-specs/pull/3338/files

jimmygchen:
Hi @ralexstokes 
I've updated this PR to the latest spec - please take another look 🙏 
Cheers

ralexstokes (approved):
looks great!

I see what appear to be unnecessary generics on some of the blob types (e.g. params like `BYTES_PER_BLOB`)

that being said, I am doing some more extensive refactoring and want to just go ahead and merge this in to avoid excessive thrash from repo-wide change

if you run into some problem please feel free to open another PR

Issue #185, opened 2023-02-09
Implement networking module across forks


brech1:
## Description

The current [networking.rs](https://github.com/ralexstokes/ethereum-consensus/blob/main/src/networking.rs) module only implements the [phase0](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md) p2p interface specs.

Since the general approach is to duplicate the definitions across spec versions, a proposed solution is to make a new networking module for each fork with the correct types.

ralexstokes:
fixed in #186

PR #186, opened 2023-02-09
Implement networking module across forks

brech1:
## Description

Solves #185 

This PR adds the specific types for the p2p networking interface for altair, bellatrix and capella. 

## Changes

- Place the `PeerId` implementation of the `networking.rs` file in the `src` directory in a new `libp2p.rs` file.
- Add the current networking implementation to the phase0 module
- Implement the updates in the specs for altair, bellatrix and capella under each module.


src/libp2p.rs
```diff
@@ -1,23 +1,10 @@
-use crate::primitives::Epoch;
```
ralexstokes, line 1: nitpick but I'd just leave this as `networking.rs` in the top-level

src/libp2p.rs
```diff
@@ -130,20 +117,6 @@ impl FromStr for PeerId {
     }
 }
 
-pub type Enr = enr::Enr<enr::k256::ecdsa::SigningKey>;
```
ralexstokes, line 133: I don't see a reason to duplicate the `Enr` type in each spec, it can just live here

ralexstokes: same thing w/ `MessageDomain`

src/phase0/networking.rs
```diff
@@ -0,0 +1,29 @@
+use crate::primitives::Epoch;
+use enr;
+pub use multiaddr::Multiaddr;
+#[cfg(feature = "serde")]
+use ssz_rs::prelude::Bitvector;
+use std::time::Duration;
```
ralexstokes, line 6: whitespace after imports, I *think* `rustfmt` will do this for you...

src/altair/networking.rs
```diff
@@ -0,0 +1,31 @@
...
+pub const ATTESTATION_SUBNET_COUNT: usize = 64;
+pub const SYNC_COMMITTEE_SUBNET_COUNT: usize = 4;
+pub const GOSSIP_MAX_SIZE: usize = 2usize.pow(20);
+pub const MAX_REQUEST_BLOCKS: usize = 2usize.pow(10);
+pub const MIN_EPOCHS_FOR_BLOCK_REQUESTS: Epoch = 33024;
+pub const MAX_CHUNK_SIZE: usize = 2usize.pow(20);
+pub const TTFB_TIMEOUT: Duration = Duration::from_secs(5);
+pub const RESP_TIMEOUT: Duration = Duration::from_secs(10);
+pub const ATTESTATION_PROPAGATION_SLOT_RANGE: usize = 32;
+pub const MAXIMUM_GOSSIP_CLOCK_DISPARITY: Duration = Duration::from_millis(500);
```
ralexstokes, line 16: imo these can all live in `phase0/networking.rs`

if it is nicer to expose them via `altair::networking` for some reason then we can just `pub use ...` here

src/altair/validator.rs
```diff
@@ -1,8 +1,7 @@
+use crate::altair::SYNC_COMMITTEE_SUBNET_COUNT;
 use crate::primitives::{BlsSignature, Root, Slot, ValidatorIndex};
 use ssz_rs::prelude::*;
 
-pub const SYNC_COMMITTEE_SUBNET_COUNT: usize = 4;
```
ralexstokes, line 4: id keep this here to mirror the `consensus-specs` and then import in the other place

src/bellatrix/networking.rs
```diff
@@ -0,0 +1,33 @@
...
+    ValidSnappy,
+}
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct MetaData {
+    #[serde(with = "crate::serde::as_string")]
+    pub seq_number: u64,
+    pub attnets: Bitvector<ATTESTATION_SUBNET_COUNT>,
+    pub syncnets: Bitvector<SYNC_COMMITTEE_SUBNET_COUNT>,
+}
```
ralexstokes, line 33: we don't need to define this, can just `pub use crate::altair::networking::MetaData` if needed

src/capella/networking.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub type Enr = enr::Enr<enr::k256::ecdsa::SigningKey>;
+
+pub enum MessageDomain {
+    InvalidSnappy,
+    ValidSnappy,
+}
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct MetaData {
```
ralexstokes, line 28: same here, can even drop `capella` and `bellatrix` files

ralexstokes (commented):
thanks for this!

I left some suggestions to simplify things -- I think we can get away with just defining the `phase0` and `altair` `networking` modules

src/libp2p.rs
brech1: Agreed

src/capella/networking.rs
brech1: Yes, however I kept the `bellatrix` module to expose the new `GOSSIP_MAX_SIZE_BELLATRIX` and `MAX_CHUNK_SIZE_BELLATRIX` constants.

src/bellatrix/networking.rs
brech1: Agreed

src/phase0/networking.rs
brech1: Sure, I thought this was on purpose. Apparently if there is already some whitespace `rustfmt` will default to a single whitespace. But if there is none, it won't do anything (tried with `const` and `fn` after imports).

src/libp2p.rs
brech1: Yes, done!

brech1:
@ralexstokes thanks for your review. Based on it the PR changes are:

- Kept the `networking.rs` file and move the specific networking types to the `phase0` networking module.
- Add the `altair` and `bellatrix` modules with only the new specified types/constants.
- I didn't expose the types cross-module at all (like doing `pub use crate::altair::networking::MetaData` on the new ones). I think this is ok, but if you think it would be nicer let me know.

src/phase0/networking.rs
ralexstokes: no, it was an omission for sure, thanks for adding it!

src/phase0/networking.rs
```diff
@@ -0,0 +1,21 @@
+use crate::primitives::Epoch;
+#[cfg(feature = "serde")]
```
ralexstokes, line 2:
```suggestion
```

src/altair/networking.rs
```diff
@@ -0,0 +1,12 @@
+use super::SYNC_COMMITTEE_SUBNET_COUNT;
```
ralexstokes, line 1: nitpick
```suggestion
use crate::altair::validator::SYNC_COMMITTEE_SUBNET_COUNT;
```

src/altair/networking.rs
```diff
@@ -0,0 +1,12 @@
...
+use crate::phase0::ATTESTATION_SUBNET_COUNT;
+#[cfg(feature = "serde")]
```
ralexstokes, line 3: looks like this was accidentally copied, and will also fail the "no default features" build (cf. #191 )
```suggestion
```

ralexstokes: the same thing is below where I also left a suggestion

ralexstokes (commented):
nice work! a few comments to resolve and then this is ready to merge :)

src/altair/networking.rs
brech1: I removed both lines, thanks!
However, related to #191, shouldn't I also add the `#[cfg_attr(feature = "serde", serde(with = "crate::serde::as_string"))]` in the Metadata struct? Or would that be managed on that PR. Let me know, thanks again.

src/phase0/networking.rs
brech1: Thanks!

src/altair/networking.rs
brech1: Missed that, thanks!

src/altair/networking.rs
ralexstokes: I'd say we resolve that other there, thanks!

ralexstokes (approved):
looks good! if you can rebase from `main` then I'll merge :)

PR #167, opened 2022-12-27
started implementing no-std support

dharjeezy:
started implementing display for errors

PR #191, opened 2023-02-24
Fix building with no 'serde' feature

willemolding:
When trying to build the crate without the serde feature enabled e.g.
```
cargo check --no-default-features
```
It was failing because the field attributes were not gated by the feature, only the struct attributes.

---

This PR uses `#[cfg_attr(feature = "serde", ...)]` on the field attributes as well. This allows the project to build with `--no-default-features` again.


willemolding:
I just saw this is all included in #167. Feel free to close if it is easier to keep it all together.

ralexstokes:
went ahead and just made `serde` required in #273 
thanks for this though!

PR #194, opened 2023-03-31
Update Capella fork version and epoch

### LINKED ISSUE #190: Update `sepolia` `fork_version` values in the config and set `capella` fork epoch
^

jimmygchen:
Fixes #190 

Update values to match:
https://github.com/eth-clients/sepolia/blob/main/bepolia/config.yaml#L32-L33


ralexstokes (approved):
excellent thanks!

PR #225, opened 2023-09-07
deneb: implement poly comm spec

### LINKED ISSUE #224: deneb: implement poly comm spec
assigned @EchoAlice

https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md

EchoAlice:
Fixes #224


ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,94 @@
...
+    let commit = c_kzg::KzgCommitment::blob_to_kzg_commitment(&blob, kzg_settings)?;
+
+    let bytes_commit = commit.to_bytes();
+    let bytevector_commit = ByteVector::try_from(bytes_commit.as_ref()).unwrap();
+
+    Ok(KzgCommitment(bytevector_commit))
+}
+
+/// Compute KZG proof at point 'z' for the polynomial represented by 'blob'.
+/// Do this by computing the qoutient polynomial in evaluation form: q(x) = (p(x) - p(z)) / (x - z).
```
ralexstokes, line 55:
```suggestion
/// Do this by computing the quotient polynomial in evaluation form: q(x) = (p(x) - p(z)) / (x - z).
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,94 @@
...
+    // Inner: g1_lincomb(bit_reversal_permutation(KZG_SETUP_LAGRANGE), blob_to_polynomial(blob)) -> KzgCommitment
+    let commit = c_kzg::KzgCommitment::blob_to_kzg_commitment(&blob, kzg_settings)?;
+
+    let bytes_commit = commit.to_bytes();
+    let bytevector_commit = ByteVector::try_from(bytes_commit.as_ref()).unwrap();
+
+    Ok(KzgCommitment(bytevector_commit))
+}
+
+/// Compute KZG proof at point 'z' for the polynomial represented by 'blob'.
```
ralexstokes, line 54: use backticks if you are referring to code elements

like _`blob`_ instead of _'blob'_

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Do this by computing the qoutient polynomial in evaluation form: q(x) = (p(x) - p(z)) / (x - z).
+fn compute_kzg_proof(
+    blob: Blob,
+    z_bytes: Bytes32,
+    kzg_settings: &KzgSettings,
+) -> Result<(KzgProof, Bytes32), Error> {
```
ralexstokes, line 60: what is this other random `Bytes32`? add some docs?

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,94 @@
+#![allow(unused)]
+use crate::{crypto::hash, ssz::ByteVector};
+use alloy_primitives::{uint, U256};
+use blst::min_pk::PublicKey;
+pub use c_kzg::{Bytes32, Bytes48, Error, KzgSettings};
```
ralexstokes, line 5: `pub` here makes everything you import in this line visible.. is this what you want?

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,94 @@
...
+    let bytes = blob.0.as_ref();
+    let blob = c_kzg::Blob::from_bytes(bytes).unwrap();
+
+    // Inner: g1_lincomb(bit_reversal_permutation(KZG_SETUP_LAGRANGE), blob_to_polynomial(blob)) -> KzgCommitment
+    let commit = c_kzg::KzgCommitment::blob_to_kzg_commitment(&blob, kzg_settings)?;
+
+    let bytes_commit = commit.to_bytes();
+    let bytevector_commit = ByteVector::try_from(bytes_commit.as_ref()).unwrap();
+
+    Ok(KzgCommitment(bytevector_commit))
```
ralexstokes, line 51: this starts to get nitpicky but I'd consider different variable names

you seem to like using the types for names but this is redundant information bc the compiler can always tell you the type

I also personally find the comments confusing as they are just leaking details of the functions you are using..

putting this all together, I'd do something like this:

```suggestion
    let inner = &blob.0;
    let blob = c_kzg::Blob::from_bytes(inner.as_ref()).unwrap();
    
    let commitment = c_kzg::KzgCommitment::blob_to_kzg_commitment(&blob, kzg_settings)?;
   
    let inner = ByteVector::try_from(commitment.to_bytes().as_slice()).unwrap();
    Ok(KzgCommitment(inner))
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: Nope

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: The function returns the combined proof and evaluation of the polynomial.  Added docs

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,92 @@
...
+
+#[derive(SimpleSerialize, Default, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct KzgCommitment(ByteVector<BYTES_PER_COMMITMENT>);
+
+#[derive(SimpleSerialize, Default, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct KzgProof(ByteVector<BYTES_PER_PROOF>);
+
+/// Uses multicalar multiplication to combine trusted setup with blob
```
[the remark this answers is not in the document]
ralexstokes: this being said, these read like they are notes to the author

given we trust our internals to work well, this leaks a bunch of details to the user of this function `blob_to_kzg_commitment`

and honestly I think the name / types are clear enough that we don't really need any further description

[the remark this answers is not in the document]
EchoAlice: That makes sense.  Thanks

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,92 @@
...
+    let commitment = c_kzg::KzgCommitment::blob_to_kzg_commitment(&blob, kzg_settings)?;
+    let inner = ByteVector::try_from(commitment.to_bytes().as_slice()).unwrap();
+
+    Ok(KzgCommitment(inner))
+}
+
+/// Compute KZG proof at point `z` for the polynomial represented by 'blob'.
+/// Do this by computing the quotient polynomial in evaluation form: q(x) = (p(x) - p(z)) / (x - z).
+/// Returns combined proof and the evaluation of the polynomial.
+fn compute_kzg_proof(
```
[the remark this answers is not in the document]
ralexstokes: maybe like `compute_kzg_proof_and_evaluation` although im not sure how much you want to stick strictly to the spec -- there is enough value in spec conformity to override the lack of semantics here

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,138 @@
+#![allow(unused)]
+use crate::{crypto::hash, ssz::ByteVector};
+use alloy_primitives::{uint, U256};
+use blst::min_pk::PublicKey;
+use c_kzg::{Bytes32, Bytes48, Error, KzgSettings};
+use ssz_rs::prelude::*;
```
ralexstokes, line 6: btw once you rebase from `main`, you can do this

```suggestion
use crate::{crypto::hash, ssz::prelude::*};
use alloy_primitives::{uint, U256};
use blst::min_pk::PublicKey;
use c_kzg::{Bytes32, Bytes48, Error, KzgSettings};
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,138 @@
+#![allow(unused)]
+use crate::{crypto::hash, ssz::ByteVector};
+use alloy_primitives::{uint, U256};
+use blst::min_pk::PublicKey;
```
ralexstokes, line 4: why is this here?

imo you'll save a lot of time/confusion if you _don't_ develop with things like `#![allow(unused)]`

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,138 @@
...
+    commitments_bytes: &[Bytes48],
+    proofs_bytes: &[Bytes48],
+    kzg_settings: &KzgSettings,
+) -> Result<bool, Error> {
+    let mut c_kzg_blobs = Vec::with_capacity(blobs.len());
+
+    for bytes in blobs.iter().map(|blob| blob.0.as_ref()) {
+        let blob = c_kzg::Blob::from_bytes(bytes)?;
+        c_kzg_blobs.push(blob);
+    }
```
ralexstokes, line 128: this works but a bit of a preference to keep logic out of the iterator, just for comprehensibility/readability

```suggestion
    for bytes in blobs {
        let inner = blob.0.as_ref();
        let blob = c_kzg::Blob::from_bytes(inner)?;
        c_kzg_blobs.push(blob);
    }
```

ralexstokes: another option if you want to get functional, something like this should work

```rust
let c_zkg_blobs = blobs.iter().map(|blob| c_kzg::Blob::from_bytes(blob.0.as_ref())).collect<Result<Vec<_>,_>()?;
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    let out = c_kzg::KzgProof::verify_blob_kzg_proof_batch(
+        &c_kzg_blobs,
+        commitments_bytes,
+        proofs_bytes,
+        kzg_settings,
+    )?;
+
+    Ok(out)
```
ralexstokes, line 137: more of an FYI but you can also do stuff like this
```suggestion
    c_kzg::KzgProof::verify_blob_kzg_proof_batch(
        &c_kzg_blobs,
        commitments_bytes,
        proofs_bytes,
        kzg_settings,
    ).map_err(Into::into)
```

ralexstokes (commented):
looks good! left some comments

ethereum-consensus/src/deneb/polynomial_commitments.rs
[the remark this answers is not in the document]
EchoAlice: Thinking that the ProofAndEvaluation return type makes a lot of sense... Doesn't feel like it's really adding much complexity and allows for us to stick to the spec.

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: i've just got a lot of warnings that are irrelevant

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: Went with the more readable option

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,148 @@
+use crate::{primitives, ssz::prelude::*};
+use alloy_primitives::{uint, U256};
+pub use c_kzg::KzgSettings;
+use c_kzg::{Bytes32, Bytes48, Error};
+use std::ops::Deref;
+
+pub const BLS_MODULUS: U256 =
+    uint!(52435875175126190479447740508185965837690552500527637822603658699938581184513_U256);
+pub const BYTES_PER_BLOB: usize = 32 * 4096;
```
ralexstokes, line 9: delete bc this is in the preset

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,148 @@
...
+pub const BYTES_PER_CONTEXT: usize = 10;
+pub const BYTES_PER_COMMITMENT: usize = 48;
+pub const BYTES_PER_FIELD_ELEMENT: usize = 32;
+pub const BYTES_PER_PROOF: usize = 48;
+pub const KZG_COMMITMENT_BYTES_LEN: usize = 48;
+pub const KZG_PROOF_BYTES_LEN: usize = 48;
+
+pub type VersionedHash = primitives::Bytes32;
```
ralexstokes, line 17:
```suggestion
pub type VersionedHash = Bytes32;
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,148 @@
+use crate::{primitives, ssz::prelude::*};
```
ralexstokes, line 1:
```suggestion
use crate::{primitives::Bytes32, ssz::prelude::*};
```

ralexstokes (changes requested):
looks good! left some comments for the direction to move in

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,148 @@
...
+pub const BYTES_PER_FIELD_ELEMENT: usize = 32;
+pub const BYTES_PER_PROOF: usize = 48;
+pub const KZG_COMMITMENT_BYTES_LEN: usize = 48;
+pub const KZG_PROOF_BYTES_LEN: usize = 48;
+
+pub type VersionedHash = primitives::Bytes32;
+pub type BLSFieldElement = U256;
+pub type Polynomial = Vec<BLSFieldElement>; // Should this polynomial type be an array?
+
+pub struct Blob(ByteVector<BYTES_PER_BLOB>);
```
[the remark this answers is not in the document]
EchoAlice: No need to. Ending up using an already-defined blob type.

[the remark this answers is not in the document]
EchoAlice: > `BYTES_PER_BLOB` should be const param

No need to. Ending up using an already-defined blob type.

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,148 @@
...
+    type Target = ByteVector<BYTES_PER_COMMITMENT>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+#[derive(SimpleSerialize, Default, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct KzgProof(ByteVector<BYTES_PER_PROOF>);
```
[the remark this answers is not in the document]
EchoAlice: Error:  the trait `ssz_rs::Serializable` is not implemented for `Bytes48`

ethereum-consensus/Cargo.toml
```diff
@@ -42,6 +43,7 @@ multihash = { version = "0.16", default-features = false, features = [
     "sha2",
 ] }
 multiaddr = "0.14.0"
+c-kzg = { git = "https://github.com/ethereum/c-kzg-4844", rev = "f5f6f863d475847876a2bd5ee252058d37c3a15d" }
```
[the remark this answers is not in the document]
EchoAlice: We can!

ethereum-consensus/src/deneb/polynomial_commitments.rs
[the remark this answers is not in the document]
ralexstokes: where do we need this for `Bytes48`?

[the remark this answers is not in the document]
EchoAlice: Not sure... But i got the type aliases for Commitment and Proof working after implementing my own type alias' 
`pub type KzgCommitment = Bytes48;`
`pub type KzgProof = Bytes48;`
 with a custom `Bytes48`.
 `pub type Bytes48 = ByteVector<48>;`

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,148 @@
...
+    Ok(KzgProof(proof))
+}
+
+pub fn verify_kzg_proof(
+    commitment_bytes: Bytes48,
+    z_bytes: Bytes32,
+    y_bytes: Bytes32,
+    proof_bytes: Bytes48,
+    kzg_settings: &KzgSettings,
+) -> Result<bool, Error> {
```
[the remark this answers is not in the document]
EchoAlice: Done.  Could you verify that the errors i implemented are correct?

https://github.com/ralexstokes/ethereum-consensus/pull/225/commits/c63cb00c4f5706ef768f8f54e319168554665f93

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
+use crate::{deneb::blob_sidecar::Blob, primitives::Bytes32, ssz::prelude::*};
+use c_kzg::Error;
+pub use c_kzg::KzgSettings;
+
+use super::mainnet::FIELD_ELEMENTS_PER_BLOB;
```
ralexstokes, line 5: this should not be here, this module is parametric over the preset constants

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+
+pub const BYTES_PER_FIELD_ELEMENT: usize = 32;
+pub const BYTES_PER_BLOB: usize = BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB;
```
ralexstokes, line 8: same here, this is a parameter of this module so we shouldn't define a concrete instantiation of it

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+pub struct ProofAndEvaluation {
+    pub proof: KzgProof,
+    pub evaluation: Bytes32,
+}
+
+pub fn blob_to_kzg_commitment<const BYTES_PER_BLOB: usize>(
+    blob: &Blob<BYTES_PER_BLOB>,
+    kzg_settings: &KzgSettings,
+) -> Result<KzgCommitment, Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref()).unwrap();
```
ralexstokes, line 27: what happens if `from_bytes` fails? this `unwrap` would panic

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    let commitment = c_kzg::KzgCommitment::blob_to_kzg_commitment(&blob, kzg_settings)?;
+    let inner = ByteVector::try_from(commitment.to_bytes().as_slice()).unwrap();
```
ralexstokes, line 30: this use of `unwrap` should be fine but I'd consider `expect("correct size")` or something to indicate that we in fact expect this to have no error

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+    Ok(inner)
+}
+
+pub fn compute_kzg_proof<const BYTES_PER_BLOB: usize>(
+    blob: &Blob<BYTES_PER_BLOB>,
+    evaluation_point: &FieldElement,
+    kzg_settings: &KzgSettings,
+) -> Result<ProofAndEvaluation, Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref()).unwrap();
```
ralexstokes, line 39: will `from_bytes` alway succeed here?

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+    let evaluation_point = c_kzg::Bytes32::from_bytes(evaluation_point.as_ref())?;
+
+    let (proof, evaluation) =
+        c_kzg::KzgProof::compute_kzg_proof(&blob, &evaluation_point, kzg_settings)?;
+    let proof = ByteVector::try_from(proof.to_bytes().as_ref()).unwrap();
+    let evaluation = ByteVector::try_from(evaluation.as_slice()).unwrap();
```
ralexstokes, line 45: you can use `expect` if you want to more clearly communicate to readers that you know there will be no error here

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+    kzg_settings: &KzgSettings,
+) -> Result<KzgProof, Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref()).unwrap();
+    let commitment = c_kzg::Bytes48::from_bytes(commitment.as_ref()).unwrap();
+
+    let ckzg_proof = c_kzg::KzgProof::compute_blob_kzg_proof(&blob, &commitment, kzg_settings)?;
+
+    let bytes_proof = ckzg_proof.to_bytes();
+    let proof = ByteVector::try_from(bytes_proof.as_ref()).unwrap();
+    Ok(proof)
```
ralexstokes, line 63: this is more stylistic but we can be a bit more compact

```rust
    let proof = c_kzg::KzgProof::compute_blob_kzg_proof(&blob, &commitment, kzg_settings)?;
    Ok(ByteVector::try_from(proof.to_bytes().as_ref()).unwrap())
```

another thing is that we aren't using our type alias we went to all the trouble to define, so I'd argue the intent of this code is clearer if we do this

```rust
    let proof = c_kzg::KzgProof::compute_blob_kzg_proof(&blob, &commitment, kzg_settings)?;
    Ok(KzgProof::try_from(proof.to_bytes().as_ref()).unwrap())
```

and we can actually go one step further using `TryInto` and some type inference

```rust
Ok(proof.to_bytes().try_into().expect("input is correct size"))
```

ethereum-consensus/src/deneb/polynomial_commitments.rs:61:49
ralexstokes, line 1: what is this?

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+    let proof = c_kzg::Bytes48::from_bytes(proof.as_ref()).unwrap();
+
+    let res = c_kzg::KzgProof::verify_kzg_proof(
+        &commitment,
+        &evaluation_point,
+        &result_point,
+        &proof,
+        kzg_settings,
+    )
+    .map_err(Into::into);
```
ralexstokes, line 85: I don't see the code that extends the error type, but we should be able to wrap the `c_kzg::Error` type naturally so that we can just use `?`
```suggestion
    )?;
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    if res? == true {
+        Ok(())
+    } else {
+        Err(Error::InvalidKzgProof(String::from("Invalid Proof")))
+    }
```
ralexstokes, line 91: this works, but I'll leave an example leveraging some of rust's more functional APIs

```suggestion
    res.then_some(()).ok_or(Error::InvalidKzgProof(..))
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+    commitments: &[KzgCommitment],
+    proofs: &[KzgProof],
+    kzg_settings: &KzgSettings,
+) -> Result<(), Error> {
+    let mut c_kzg_blobs = Vec::with_capacity(blobs.len());
+    let mut c_kzg_commitments = Vec::with_capacity(commitments.len());
+    let mut c_kzg_proofs = Vec::with_capacity(proofs.len());
+
+    // TODO: Can we assume there are *always* the same number of
+    //       blobs, commitments and proofs?  We could then use one for loop.
```
ralexstokes, line 125: what does the c-kzg API expect?

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+pub use c_kzg::KzgSettings;
+
+use super::mainnet::FIELD_ELEMENTS_PER_BLOB;
+
+pub const BYTES_PER_FIELD_ELEMENT: usize = 32;
+pub const BYTES_PER_BLOB: usize = BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB;
+pub const BYTES_PER_COMMITMENT: usize = 48;
+pub const BYTES_PER_PROOF: usize = 48;
+
+pub type Bytes48 = ByteVector<48>;
```
ralexstokes, line 12: I don't see a reason to export this `Bytes48` type

and instead I'd prolly just do `pub type Kzg{Commitment, Proof} = ByteVector<48>;`

ralexstokes (changes requested):
great updates :)

ethereum-consensus/src/deneb/polynomial_commitments.rs
ralexstokes: use the constant as the length

ethereum-consensus/src/deneb/polynomial_commitments.rs
ralexstokes: ?

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,152 @@
...
+pub type VersionedHash = Bytes32;
+pub type FieldElement = Bytes32;
+pub type KzgCommitment = Bytes48;
+pub type KzgProof = Bytes48;
+
```
ralexstokes, line 17: should define a local Error type here...

something like

```rust
pub struct Error(c_kzg::Error),
```

ralexstokes: also add a variant in `crate::state_transition::Error` that wraps this one

```rust
#[error(transparent)]
PolynomialCommitment(crate::deneb::polynomial_commitment::Error),
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: Idk how to implement From for it yet.  Sticking with ` Ok(KzgProof::try_from(proof.to_bytes().as_ref()).expect("input is correct size"))` for now

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: What exactly is this?   
`pub type Kzg{Commitment, Proof} = ByteVector<48>;`

ralexstokes: just a shorthand so I don't have to type as much


https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html

ethereum-consensus/Cargo.toml
```diff
@@ -16,8 +16,8 @@ ec = [
     "bip39",
     "rand_core",
     "rayon",
-    "hkdf",
     "ruint",
+    "hkdf",
```
ralexstokes, line 20: unnecessary change but its fine

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,147 @@
...
+        // Implement custom formatting logic here
+        write!(f, "Custom error: {:?}", self.0)
+    }
+}
+
+impl From<c_kzg::Error> for Error {
+    fn from(inner_error: c_kzg::Error) -> Self {
+        Error(inner_error)
+    }
+}
```
ralexstokes, line 29: there's a shortcut here w/ `thiserror`
```suggestion
pub struct Error(#[from] c_kzg::Error);

impl fmt::Display for Error {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        // Implement custom formatting logic here
        write!(f, "Custom error: {:?}", self.0)
    }
}
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,147 @@
...
+
+pub fn blob_to_kzg_commitment<const BYTES_PER_BLOB: usize>(
+    blob: &Blob<BYTES_PER_BLOB>,
+    kzg_settings: &KzgSettings,
+) -> Result<KzgCommitment, Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref())?;
+
+    let commitment = c_kzg::KzgCommitment::blob_to_kzg_commitment(&blob, kzg_settings)?;
+    let inner = ByteVector::try_from(commitment.to_bytes().as_slice()).expect("Correct size");
```
ralexstokes, line 43: lowercase better here
```suggestion
    let inner = ByteVector::try_from(commitment.to_bytes().as_slice()).expect("correct size");
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,147 @@
...
+    evaluation_point: &FieldElement,
+    kzg_settings: &KzgSettings,
+) -> Result<ProofAndEvaluation, Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref())?;
+    let evaluation_point = c_kzg::Bytes32::from_bytes(evaluation_point.as_ref())?;
+
+    let (proof, evaluation) =
+        c_kzg::KzgProof::compute_kzg_proof(&blob, &evaluation_point, kzg_settings)?;
+    let proof = ByteVector::try_from(proof.to_bytes().as_ref()).expect("Correct size");
+    let evaluation = ByteVector::try_from(evaluation.as_slice()).expect("Correct size");
```
ralexstokes, line 58:
```suggestion
    let proof = ByteVector::try_from(proof.to_bytes().as_ref()).expect("correct size");
    let evaluation = ByteVector::try_from(evaluation.as_slice()).expect("correct size");
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,147 @@
...
+    Ok(result)
+}
+
+pub fn compute_blob_kzg_proof<const BYTES_PER_BLOB: usize>(
+    blob: &Blob<BYTES_PER_BLOB>,
+    commitment: &KzgCommitment,
+    kzg_settings: &KzgSettings,
+) -> Result<KzgProof, Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref())?;
+    let commitment = c_kzg::Bytes48::from_bytes(commitment.as_ref()).expect("Correct size");
```
ralexstokes, line 70:
```suggestion
    let commitment = c_kzg::Bytes48::from_bytes(commitment.as_ref()).expect("correct size");
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,147 @@
...
+    commitment: &KzgCommitment,
+    evaluation_point: &FieldElement,
+    result_point: &FieldElement,
+    proof: &KzgProof,
+    kzg_settings: &KzgSettings,
+) -> Result<(), Error> {
+    let evaluation_point = c_kzg::Bytes32::from_bytes(evaluation_point.as_ref())?;
+    let result_point = c_kzg::Bytes32::from_bytes(result_point.as_ref())?;
+    let commitment = c_kzg::Bytes48::from_bytes(commitment.as_ref()).expect("Correct size");
+    let proof = c_kzg::Bytes48::from_bytes(proof.as_ref()).expect("Correct size");
```
ralexstokes, line 87:
```suggestion
    let commitment = c_kzg::Bytes48::from_bytes(commitment.as_ref()).expect("correct size");
    let proof = c_kzg::Bytes48::from_bytes(proof.as_ref()).expect("correct size");
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,147 @@
...
+    res.then_some(()).ok_or(Error(c_kzg::Error::InvalidKzgProof(String::from("Invalid Proof"))))
+}
+
+pub fn verify_blob_kzg_proof<const BYTES_PER_BLOB: usize>(
+    blob: &Blob<BYTES_PER_BLOB>,
+    commitment: &KzgCommitment,
+    proof: &KzgProof,
+    kzg_settings: &KzgSettings,
+) -> Result<(), Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref()).unwrap();
```
ralexstokes, line 106: we don't know that our `Blob` type matches the size the `c_kzg::Blob` expects so we should not `unwrap` the error blindly
```suggestion
    let blob = c_kzg::Blob::from_bytes(blob.as_ref())?;
```

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -55,6 +55,8 @@ pub enum Error {
     UnknownPreset(String),
     #[error(transparent)]
     ExecutionEngine(#[from] ExecutionEngineError),
+    #[error(transparent)]
+    PolynomialCommitment(#[from] crate::deneb::polynomial_commitments::Error),
```
ralexstokes, line 59: easier to read if you do the import at the top and just bind some local symbol here

```rust
use crate::deneb::polynomial_commitments::Error as PolynomialCommitmentError;
```

and then just 
```suggestion
    PolynomialCommitment(#[from] PolynomialCommitmentError),
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,147 @@
...
+
+#[derive(Debug, Error)]
+pub struct Error(c_kzg::Error);
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // Implement custom formatting logic here
+        write!(f, "Custom error: {:?}", self.0)
+    }
+}
```
ralexstokes, line 23: I think we can just drop this?

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    let res = c_kzg::KzgProof::verify_kzg_proof(
+        &commitment,
+        &evaluation_point,
+        &result_point,
+        &proof,
+        kzg_settings,
+    )?;
+
+    res.then_some(()).ok_or(Error(c_kzg::Error::InvalidKzgProof(String::from("Invalid Proof"))))
```
ralexstokes, line 97: if you go see how these are used in the code, these variants `InvalidKzg{Proof, Commitment}` are just used to signal that the size of the input data is the wrong

so two things here:
1) you are overloading the meaning of this term in our code, which is likely to confuse other readers (including yourself in the future)
2) we also limit the scope of how much the `c_kzg` types leak out of this module by not reusing them to forward on to others

to fix these:

let's extend the `Error` type you have above to an enum with variants for all of our possible errors

```rust
enum Error {
    #[error(transparent)]
    CKzg(c_kzg::Error),
    #[error("proof verification failed")]
    InvalidProof,
}
```

ralexstokes (changes requested):
almost there!

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: `error[E0277]: `polynomial_commitments::Error` doesn't implement `std::fmt::Display``

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: `Error::InvalidProof` suffices for all 3 functions?

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,137 @@
...
+
+pub type VersionedHash = Bytes32;
+pub type FieldElement = Bytes32;
+pub type KzgCommitment = ByteVector<BYTES_PER_COMMITMENT>;
+pub type KzgProof = ByteVector<BYTES_PER_PROOF>;
+
+#[derive(Debug, Error)]
+#[error(transparent)]
+pub enum Error {
+    CKzg(#[from] c_kzg::Error),
```
ralexstokes, line 17:
```suggestion
pub enum Error {
    #[error(transparent)]
    CKzg(#[from] c_kzg::Error),
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[error("proof verification failed")]
+    InvalidProof,
+}
+
+pub struct ProofAndEvaluation {
+    pub proof: KzgProof,
+    pub evaluation: Bytes32,
```
ralexstokes, line 24:
```suggestion
    pub evaluation: FieldElement,
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+pub fn blob_to_kzg_commitment<const BYTES_PER_BLOB: usize>(
+    blob: &Blob<BYTES_PER_BLOB>,
+    kzg_settings: &KzgSettings,
+) -> Result<KzgCommitment, Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref())?;
+
+    let commitment = c_kzg::KzgCommitment::blob_to_kzg_commitment(&blob, kzg_settings)?;
+    let inner = ByteVector::try_from(commitment.to_bytes().as_slice()).expect("correct size");
```
ralexstokes, line 34:
```suggestion
    let inner = KzgCommitment::try_from(commitment.to_bytes().as_slice()).expect("correct size");
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,137 @@
...
+    evaluation_point: &FieldElement,
+    kzg_settings: &KzgSettings,
+) -> Result<ProofAndEvaluation, Error> {
+    let blob = c_kzg::Blob::from_bytes(blob.as_ref())?;
+    let evaluation_point = c_kzg::Bytes32::from_bytes(evaluation_point.as_ref())?;
+
+    let (proof, evaluation) =
+        c_kzg::KzgProof::compute_kzg_proof(&blob, &evaluation_point, kzg_settings)?;
+    let proof = ByteVector::try_from(proof.to_bytes().as_ref()).expect("correct size");
+    let evaluation = ByteVector::try_from(evaluation.as_slice()).expect("correct size");
```
ralexstokes, line 49:
```suggestion
    let proof = KzgProof::try_from(proof.to_bytes().as_ref()).expect("correct size");
    let evaluation = FieldElement::try_from(evaluation.as_slice()).expect("correct size");
```

ethereum-consensus/src/deneb/polynomial_commitments.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    let res = c_kzg::KzgProof::verify_kzg_proof(
+        &commitment,
+        &evaluation_point,
+        &result_point,
+        &proof,
+        kzg_settings,
+    )?;
+
+    res.then_some(()).ok_or(Error::InvalidProof)
```
ralexstokes, line 88: just leaving a note to show how this more "functional" style leads to a nice "fluent" approach ("method chaining") and can result in more compact code (if that's desired)

```rust
c_kzg::KzgProof::verify_kzg_proof(..)?.then_some(()).ok_or(Error::InvalidProof)
```

ralexstokes (approved):
excellent work!

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice: For sure.  Thanks

ethereum-consensus/src/deneb/polynomial_commitments.rs
EchoAlice, line 1: I'll try to be less sloppy next time

PR #230, opened 2023-09-13
feat: add network type

### LINKED ISSUE #227: stronger typing for `Network`s concept
I started a module to house this here: https://github.com/ralexstokes/ethereum-consensus/blob/main/ethereum-consensus/src/networks.rs

it would be nice to have stronger typing for the networks we support

e.g.

```rust
enum Network {
   Mainnet,
   Goerli,
   ...
}
```

and use this in lieu of the string name in the config (see here: https://github.com/ralexstokes/ethereum-consensus/blob/main/ethereum-consensus/src/state_transition/context.rs#L75 )

we can then also refactor stuff like this: https://github.com/ralexstokes/mev-rs/blob/main/mev-rs/src/network.rs

chirag-bgh:
Closes #227

### DIFF
--- a/ethereum-consensus/src/networks.rs
+++ b/ethereum-consensus/src/networks.rs
@@ -1,5 +1,43 @@
 /// This module contains support for various Ethereum netowrks.
-use crate::state_transition::Context;
+use crate::state_transition::{Context, Error};
+
+#[derive(Default, Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "lowercase"))]
+pub enum Network {
+    #[default]
+    Mainnet,
+    Sepolia,
+    Goerli,
+    Holesky,
+    Custom(String),
+}
+
+impl std::fmt::Display for Network {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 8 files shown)


ethereum-consensus/src/networks.rs
```diff
@@ -1,5 +1,43 @@
...
+}
+
+impl std::fmt::Display for Network {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Mainnet => write!(f, "mainnet"),
+            Self::Sepolia => write!(f, "sepolia"),
+            Self::Goerli => write!(f, "goerli"),
+            Self::Holesky => write!(f, "holesky"),
+            Self::Custom(config) => write!(f, "custom network with config at `{config}`"),
```
ralexstokes, line 23: this was written for a config file, let's handle that specific concern elsewhere and for now just keep `Custom` as a variant to wrap some custom network name
```suggestion
            Self::Custom(name) => write!(f, "{name}"),
```

ralexstokes (changes requested):
looks great!

one change to make this maximally re-useable

next will be to update `mev-rs` to use this new type here

ralexstokes (approved):
looks good!

I think we may need to move that `TryFrom<&Network> for Context` impl but we can do that when we refactor `mev-rs`

PR #251, opened 2023-09-18
feat: implements process bls to execution change

### LINKED ISSUE #249: `capella`: impl `process_bls_to_execution_change`
[`process_bls_to_execution_change`](https://github.com/ralexstokes/ethereum-consensus/blob/61836a64464f311a40fd3038edf774f5c239e3ea/ethereum-consensus/src/capella/block_processing.rs#L43) following: https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#new-process_bls_to_execution_change

samtvlabs:
Took a stab at this and submitting a draft PR for feedback. 

This PR closes #249

### DIFF
--- a/ethereum-consensus/src/capella/block_processing.rs
+++ b/ethereum-consensus/src/capella/block_processing.rs
@@ -1,15 +1,14 @@
 use crate::{
     capella::{
-        compute_domain, compute_signing_root, compute_timestamp_at_slot, decrease_balance,
+        compute_domain, compute_timestamp_at_slot, decrease_balance,
         get_current_epoch, get_randao_mix, is_fully_withdrawable_validator,
         is_partially_withdrawable_validator, process_attestation, process_attester_slashing,
         process_block_header, process_deposit, process_eth1_data, process_proposer_slashing,
         process_randao, process_sync_aggregate, process_voluntary_exit, BeaconBlock,
-        BeaconBlockBody, BeaconState, BlsPublicKey, DomainType, ExecutionAddress, ExecutionEngine,
+        BeaconBlockBody, BeaconState, DomainType, ExecutionAddress, ExecutionEngine,
         ExecutionPayload, ExecutionPayloadHeader, NewPayloadRequest, SignedBlsToExecutionChange,
         Withdrawal,
     },
-    crypto::{hash, verify_signature},
     primitives::{BLS_WITHDRAWAL_PREFIX, ETH1_ADDRESS_WITHDRAWAL_PREFIX},
     ssz::prelude::*,
     state_transition::{
@@ -17,6 +16,8 @@ use crate::{
         InvalidOperation, InvalidWithdrawals, Result,
     },
 };
+
+use super::verify_signed_data;
 
 pub fn process_bls_to_execution_change<
     const SLOTS_PER_HISTORICAL_ROOT: usize,
@@ -42,7 +43,7 @@ pub fn process_bls_to_execution_change<
         BYTES_PER_LOGS_BLOOM,
         MAX_EXTRA_DATA_BYTES,
     >,
-    signed_address_change: &mut SignedBlsToExecutionChange,
+    signed_address_change: &SignedBlsToExecutionChange,
     context: &Context,
 ) -> Result<()> {
     let address_change = &signed_address_change.message;
@@ -69,6 +70,7 @@ pub fn process_bls_to_execution_change<
         Some(state.genesis_validators_root),
         context,
     )?;
+    let public_key = &address_change.from_bls_public_key;
     let signed_data =
         verify_signed_data(&mut address_change.clone(), signature, public_key, domain);
     if signed_data.is_err() {
--- a/ethereum-consensus/src/capella/block_processing.rs
+++ b/ethereum-consensus/src/capella/block_processing.rs
@@ -45,17 +45,23 @@ pub fn process_bls_to_execution_change<
     signed_address_change: &mut SignedBlsToExecutionChange,
     context: &Context,
 ) -> Result<()> {
-    let address_change = signed_address_change;
+    let address_change = &signed_address_change.message;
+    let signature = &signed_address_change.signature;
 
-    assert!(address_change.message.validator_index < state.validators.len());
+    if address_change.validator_index >= state.validators.len() {
+        return Err(invalid_operation_error(InvalidOperation::ValidatorIndex(
+            address_change.validator_index,
+        )))
+    }
 
-    let validator = &mut state.validators[address_change.message.validator_index];
+    let withdrawal_credentials_prefix =
+        state.validators[address_change.validator_index].withdrawal_credentials[0];
 
-    assert!(validator.withdrawal_credentials.starts_with(&[BLS_WITHDRAWAL_PREFIX]));
-    assert!(
-        validator.withdrawal_credentials[1..]
-            == hash(address_change.message.from_bls_public_key.as_ref())[1..]
-    );
+    if withdrawal_credentials_prefix != BLS_WITHDRAWAL_PREFIX {
+        return Err(invalid_operation_error(InvalidOperation::WithdrawalCredentialsPrefix(
+            state.validators[address_change.validator_index].withdrawal_credentials[0],
+        )))
+    }
 
     let domain = compute_domain(
         DomainType::BlsToExecutionChange,
@@ -63,15 +69,24 @@ pub fn process_bls_to_execution_change<
         Some(state.genesis_validators_root),
         context,
     )?;
-    let signing_root = compute_signing_root(address_change, domain)?;
-    let pk: &BlsPublicKey = &address_change.message.from_bls_public_key;
-    assert!(verify_signature(pk, signing_root.as_ref(), &address_change.signature,).is_ok());
-    let withdrawal_credentials = vec![ETH1_ADDRESS_WITHDRAWAL_PREFIX];
+    let signed_data =
+        verify_signed_data(&mut address_change.clone(), signature, public_key, domain);
+    if signed_data.is_err() {
+        return Err(invalid_operation_error(InvalidOperation::ExecutionChange(
+            signed_address_change.signature.clone(),
+        )))
+    }
 
-    let withdrawal_credentials_array: [u8; 32] =
-        withdrawal_credentials.try_into().expect("Wrong size");
-    validator.withdrawal_credentials = ByteVector::try_from(withdrawal_credentials_array.as_ref())
-        .expect("Failed to convert array to ByteVector");
+    let validator = &mut state.validators[address_change.validator_index];
+
+    validator.withdrawal_credentials[0] = ETH1_ADDRESS_WITHDRAWAL_PREFIX;
+    let mut withdrawal_credentials: Vec<u8> = validator.withdrawal_credentials.as_ref().to_vec();
+    withdrawal_credentials[12..].copy_from_slice(address_change.to_execution_address.as_ref());
+    let withdrawal_credentials: &[u8] = &withdrawal_credentials;
+    let withdrawal_credentials =
+        ByteVector::<32>::try_from(withdrawal_credentials).expect("Wrong size");
+    validator.withdrawal_credentials = withdrawal_credentials;
+
     Ok(())
 }
 
@@ -135,7 +150,7 @@ pub fn process_operations<
                 expected: expected_deposit_count,
                 count: body.deposits.len(),
             },
-        )));
+        )))
     }
     body.proposer_slashings
         .iter_mut()
@@ -207,7 +222,7 @@ pub fn process_execution_payload<
                 expected: state.latest_execution_payload_header.block_hash.clone(),
             }
             .into(),
-        ));
+        ))
     }
 
     let current_epoch = get_current_epoch(state, context);
@@ -219,7 +234,7 @@ pub fn process_execution_payload<
                 expected: randao_mix.clone(),
             }
             .into(),
-        ));
+        ))
     }
 
     let timestamp = compute_timestamp_at_slot(state, state.slot, context)?;
@@ -230,7 +245,7 @@ pub fn process_execution_payload<
                 expected: timestamp,
             }
             .into(),
-        ));
+        ))
     }
 
     let new_payload_request = NewPayloadRequest(payload);
--- a/ethereum-consensus/src/capella/block_processing.rs
+++ b/ethereum-consensus/src/capella/block_processing.rs
@@ -4,16 +4,17 @@ use crate::{
         get_randao_mix, is_fully_withdrawable_validator, is_partially_withdrawable_validator,
         process_attestation, process_attester_slashing, process_block_header, process_deposit,
         process_eth1_data, process_proposer_slashing, process_randao, process_sync_aggregate,
-        process_voluntary_exit, BeaconBlock, BeaconBlockBody, BeaconState, BlsPublicKey,
-        DomainType, ExecutionAddress, ExecutionEngine, ExecutionPayload, ExecutionPayloadHeader,
+        process_voluntary_exit, BeaconBlock, BeaconBlockBody, BeaconState, DomainType,
+        ExecutionAddress, ExecutionEngine, ExecutionPayload, ExecutionPayloadHeader,
         NewPayloadRequest, SignedBlsToExecutionChange, Withdrawal,
     },
+    crypto::hash,
     primitives::{BLS_WITHDRAWAL_PREFIX, ETH1_ADDRESS_WITHDRAWAL_PREFIX},
     signing::verify_signed_data,
     ssz::prelude::*,
     state_transition::{
-        invalid_operation_error, Context, InvalidDeposit, InvalidExecutionPayload,
-        InvalidOperation, InvalidWithdrawals, Result,
+        invalid_operation_error, Context, InvalidBlsToExecutionChange, InvalidDeposit,
+        InvalidExecutionPayload, InvalidOperation, InvalidWithdrawals, Result,
     },
 };
 
@@ -44,22 +45,34 @@ pub fn process_bls_to_execution_change<
     signed_address_change: &SignedBlsToExecutionChange,
     context: &Context,
 ) -> Result<()> {
-    let address_change = &signed_address_change.message;
+    let mut address_change = signed_address_change.message.clone();
     let signature = &signed_address_change.signature;
 
     if address_change.validator_index >= state.validators.len() {
-        return Err(invalid_operation_error(InvalidOperation::ValidatorIndex(
-            address_change.validator_index,
-        )));
+        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
+            InvalidBlsToExecutionChange::BadValidatorIndex(address_change.validator_index),
+        )))
     }
 
     let withdrawal_credentials_prefix =
         state.validators[address_change.validator_index].withdrawal_credentials[0];
 
     if withdrawal_credentials_prefix != BLS_WITHDRAWAL_PREFIX {
-        return Err(invalid_operation_error(InvalidOperation::WithdrawalCredentialsPrefix(
-            state.validators[address_change.validator_index].withdrawal_credentials[0],
-        )));
+        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
+            InvalidBlsToExecutionChange::WithdrawalCredentialsPrefix(
+                state.validators[address_change.validator_index].withdrawal_credentials[0],
+            ),
+        )))
+    }
+
+    let public_key = address_change.from_bls_public_key.clone();
+
+    if state.validators[address_change.validator_index].withdrawal_credentials[1..] !=
+        hash(public_key.as_slice())[1..]
+    {
+        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
+            InvalidBlsToExecutionChange::WithdrawalCredentialsPublicKey(public_key),
+        )))
     }
 
     let domain = compute_domain(
@@ -68,24 +81,15 @@ pub fn process_bls_to_execution_change<
         Some(state.genesis_validators_root),
         context,
     )?;
-    let public_key = &address_change.from_bls_public_key;
-    let signed_data =
-        verify_signed_data(&mut address_change.clone(), signature, public_key, domain);
-    if signed_data.is_err() {
-        return Err(invalid_operation_error(InvalidOperation::ExecutionChange(
-            signed_address_change.signature.clone(),
-        )));
-    }
+
+    verify_signed_data(&mut address_change, signature, &public_key, domain)?;
 
     let validator = &mut state.validators[address_change.validator_index];
 
     validator.withdrawal_credentials[0] = ETH1_ADDRESS_WITHDRAWAL_PREFIX;
-    let mut withdrawal_credentials: Vec<u8> = validator.withdrawal_credentials.as_ref().to_vec();
-    withdrawal_credentials[12..].copy_from_slice(address_change.to_execution_address.as_ref());
-    let withdrawal_credentials: &[u8] = &withdrawal_credentials;
-    let withdrawal_credentials =
-        ByteVector::<32>::try_from(withdrawal_credentials).expect("Wrong size");
-    validator.withdrawal_credentials = withdrawal_credentials;
+    validator.withdrawal_credentials[1..12].fill(0);
+    validator.withdrawal_credentials[12..]
+        .copy_from_slice(address_change.to_execution_address.as_ref());
 
     Ok(())
 }
@@ -150,7 +154,7 @@ pub fn process_operations<
                 expected: expected_deposit_count,
                 count: body.deposits.len(),
             },
-        )));
+        )))
     }
     body.proposer_slashings
         .iter_mut()
@@ -222,7 +226,7 @@ pub fn process_execution_payload<
                 expected: state.latest_execution_payload_header.block_hash.clone(),
             }
             .into(),
-        ));
+        ))
     }
 
     let current_epoch = get_current_epoch(state, context);
@@ -234,7 +238,7 @@ pub fn process_execution_payload<
                 expected: randao_mix.clone(),
             }
             .into(),
-        ));
+        ))
     }
 
     let timestamp = compute_timestamp_at_slot(state, state.slot, context)?;
@@ -245,7 +249,7 @@ pub fn process_execution_payload<
                 expected: timestamp,
             }
             .into(),
-        ));
+        ))
     }
 
     let new_payload_request = NewPayloadRequest(payload);
@@ -316,7 +320,7 @@ pub fn process_withdrawals<
                 provided: execution_payload.withdrawals.to_vec(),
                 expected: expected_withdrawals,
             },
-        )));
+        )))
     }
 
     for withdrawal in &expected_withdrawals {
@@ -401,7 +405,7 @@ pub fn get_expected_withdrawals<
             withdrawal_index += 1;
         }
         if withdrawals.len() == context.max_withdrawals_per_payload {
-            break;
+            break
         }
         validator_index = (validator_index + 1) % state.validators.len();
     }
--- a/ethereum-consensus/src/capella/block_processing.rs
+++ b/ethereum-consensus/src/capella/block_processing.rs
@@ -54,7 +54,8 @@ pub fn process_bls_to_execution_change<
         )))
     }
 
-    let withdrawal_credentials = &mut state.validators[address_change.validator_index].withdrawal_credentials;
+    let withdrawal_credentials =
+         &mut state.validators[address_change.validator_index].withdrawal_credentials;
     if withdrawal_credentials[0] != BLS_WITHDRAWAL_PREFIX {
         return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
             InvalidBlsToExecutionChange::WithdrawalCredentialsPrefix(withdrawal_credentials[0]),
--- a/ethereum-consensus/src/capella/block_processing.rs
+++ b/ethereum-consensus/src/capella/block_processing.rs
@@ -1,13 +1,16 @@
 use crate::{
     capella::{
-        compute_timestamp_at_slot, decrease_balance, get_current_epoch, get_randao_mix,
-        is_fully_withdrawable_validator, is_partially_withdrawable_validator, process_attestation,
-        process_attester_slashing, process_block_header, process_deposit, process_eth1_data,
-        process_proposer_slashing, process_randao, process_sync_aggregate, process_voluntary_exit,
-        BeaconBlock, BeaconBlockBody, BeaconState, ExecutionAddress, ExecutionEngine,
+        compute_domain, compute_signing_root, compute_timestamp_at_slot, decrease_balance,
+        get_current_epoch, get_randao_mix, is_fully_withdrawable_validator,
+        is_partially_withdrawable_validator, process_attestation, process_attester_slashing,
+        process_block_header, process_deposit, process_eth1_data, process_proposer_slashing,
+        process_randao, process_sync_aggregate, process_voluntary_exit, BeaconBlock,
+        BeaconBlockBody, BeaconState, BlsPublicKey, DomainType, ExecutionAddress, ExecutionEngine,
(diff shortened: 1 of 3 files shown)


ethereum-consensus/src/capella/bls_to_execution_change.rs
```diff
@@ -14,6 +14,6 @@ pub struct BlsToExecutionChange {
 #[derive(Default, Debug, Clone, SimpleSerialize, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct SignedBlsToExecutionChange {
-    message: BlsToExecutionChange,
-    signature: BlsSignature,
+    pub message: BlsToExecutionChange,
```
ralexstokes, line 17: nice catch, we will want these `pub` anyhow

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -103,7 +130,7 @@ pub fn process_operations<
                 expected: expected_deposit_count,
                 count: body.deposits.len(),
             },
-        )))
+        )));
```
ralexstokes, line 133: are you running the nightly `cargo`?

if you have `just` installed, you can run `just fmt` and it will execute the right command for you

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +41,33 @@ pub fn process_bls_to_execution_change<
         BYTES_PER_LOGS_BLOOM,
         MAX_EXTRA_DATA_BYTES,
     >,
-    _signed_address_change: &mut SignedBlsToExecutionChange,
-    _context: &Context,
+    signed_address_change: &mut SignedBlsToExecutionChange,
+    context: &Context,
 ) -> Result<()> {
-    unimplemented!()
+    let address_change = signed_address_change;
```
ralexstokes, line 47: to emulate the spec here, we can take a reference to the inner `message` field; then you can change all the `address_change.message` accesses to just `address_change` below

```suggestion
    let address_change = &signed_address_change.message;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +41,33 @@ pub fn process_bls_to_execution_change<
...
+
+    assert!(address_change.message.validator_index < state.validators.len());
+
+    let validator = &mut state.validators[address_change.message.validator_index];
+
+    assert!(validator.withdrawal_credentials.starts_with(&[BLS_WITHDRAWAL_PREFIX]));
```
ralexstokes, line 53: this works but I don't see a reason to not just stick more closely to the spec
```suggestion
    assert!(validator.withdrawal_credentials[0] == BLS_WITHDRAWAL_PREFIX);
```

we know the direct element access is safe as the type of `withdrawal_credentials` is always at least that big

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +41,33 @@ pub fn process_bls_to_execution_change<
...
+    assert!(
+        validator.withdrawal_credentials[1..]
+            == hash(address_change.message.from_bls_public_key.as_ref())[1..]
+    );
+
+    let domain =
+        compute_domain(BlsToExecutionChange, None, Some(state.genesis_validators_root), context)?;
```
ralexstokes, line 60: nitpick but I prefer to grab these from the enum
```suggestion
        compute_domain(DomainType::BlsToExecutionChange, None, Some(state.genesis_validators_root), context)?;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +41,33 @@ pub fn process_bls_to_execution_change<
...
+    let signing_root = compute_signing_root(address_change, domain)?;
```
ralexstokes, line 61: this needs `&mut` to be able to compute the `hash_tree_root`; also, the message being signed over is the `message` here, not the full data

```suggestion
    let signing_root = compute_signing_root(&mut address_change.message, domain)?;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +41,33 @@ pub fn process_bls_to_execution_change<
...
+    let pk: &BlsPublicKey = &address_change.message.from_bls_public_key;
```
ralexstokes, line 62: I prefer fully typing out the variable names in this repo, as it is meant to be read by many; also I don't see why we need the type annotation....

```suggestion
    let public_key = &address_change.message.from_bls_public_key;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +41,33 @@ pub fn process_bls_to_execution_change<
...
+    assert!(verify_signature(&pk, signing_root.as_ref(), &address_change.signature,).is_ok());
```
ralexstokes, line 63: double-ref here and extra comma
```suggestion
    assert!(verify_signature(pk, signing_root.as_ref(), &address_change.signature).is_ok());
```

ralexstokes: also note, if we compute the `domain`, there's a higher-level function to use: https://github.com/ralexstokes/ethereum-consensus/blob/61836a64464f311a40fd3038edf774f5c239e3ea/ethereum-consensus/src/signing.rs#L33

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +41,33 @@ pub fn process_bls_to_execution_change<
...
+    let withdrawal_credentials = vec![ETH1_ADDRESS_WITHDRAWAL_PREFIX];
+
+    let withdrawal_credentials_array: [u8; 32] =
+        withdrawal_credentials.try_into().expect("Wrong size");
+    validator.withdrawal_credentials = ByteVector::try_from(withdrawal_credentials_array.as_ref())
+        .expect("Failed to convert array to ByteVector");
```
ralexstokes, line 69: we are missing the actual address to set, and also we can work directly w/ the memory already allocated for the `withdrawal_credentials`:

```suggestion
    validator.withdrawal_credentials[0] = ETH1_ADDRESS_WITHDRAWAL_PREFIX;
    validator.withdrawal_credentials[1..12].fill(0);
    validator.withdrawal_credentials[12..].copy_from_slice(address_change.message.to_execution_address.as_ref());
```

ralexstokes (changes requested):
nice work!

there are a few changes to go ahead and make now

after those, we will want to change the use of `assert!` to errors but we can handle that after the first round of updates

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: I put changes by running the lint from the ci. Will add just

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: If we follow the spec and do

```rust
    let address_change = &signed_address_change.message;
```

we need to use the `signed_address_change` here and run into ownership issues down the line , and this consume the var.

We also can't clone it :(

ralexstokes: where is it an issue?

we should be able to separate out the message from the signature without any problems

```rust
let address_change = &mut signed_address_change.message;
let signature = &signed_address_change.signature;
```

should be fine as these are disjoint field accesses

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: I couldnt do it this way , as it yields the following error 

```shell
`ssz_rs::Vector<u8, 32>` as mutable
trait `IndexMut` is required to modify indexed content, but it is not implemented for `ssz_rs::Vector<u8, 32>`
```

ssr_rs is an external library, so we can't apply `IndexMut` to it . 

I ended up having to cast it to a vec, abd then a `ByteVector<32>`.

```rust
    validator.withdrawal_credentials[0] = ETH1_ADDRESS_WITHDRAWAL_PREFIX;
    let mut withdrawal_credentials: Vec<u8> = validator.withdrawal_credentials.as_ref().to_vec();
    withdrawal_credentials[12..].copy_from_slice(address_change.to_execution_address.as_ref());
    let withdrawal_credentials: &[u8] = &withdrawal_credentials;
    let withdrawal_credentials =
        ByteVector::<32>::try_from(withdrawal_credentials).expect("Wrong size");
    validator.withdrawal_credentials = withdrawal_credentials;
  ```

I also had a question with regards to whether or not this was redudnant since we are creating an vector now

```rust
    validator.withdrawal_credentials[1..12].fill(0);
```

I have omitted it , but happy to add it back

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: pushed with this

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: done

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: done

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: done

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: done

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -1,12 +1,15 @@
...
+        process_attestation, process_attester_slashing, process_block_header, process_deposit,
+        process_eth1_data, process_proposer_slashing, process_randao, process_sync_aggregate,
+        process_voluntary_exit, BeaconBlock, BeaconBlockBody, BeaconState, BlsPublicKey,
+        DomainType, ExecutionEngine, ExecutionPayload, ExecutionPayloadHeader, NewPayloadRequest,
+        SignedBlsToExecutionChange,
     },
-    ssz::prelude::*,
+    primitives::{BLS_WITHDRAWAL_PREFIX, ETH1_ADDRESS_WITHDRAWAL_PREFIX},
+    signing::verify_signed_data,
+    ssz::{ByteVector,prelude::*},
```
ralexstokes, line 12: went ahead and changed this so we can just do this
```suggestion
    ssz::prelude::*,
```

samtvlabs: should have checked before committing it , but ByteVector isnt inn the prelude, thats why CI is failing

ethereum-consensus/src/capella/block_processing.rs
ralexstokes: ah sorry about this, I ran into elsewhere and went ahead and patched it in `ssz_rs` here: https://github.com/ralexstokes/ssz-rs/pull/105

if you rebase from `main` you should have the updated fuctionality and should be able to do what I suggested above

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +40,54 @@ pub fn process_bls_to_execution_change<
...
+
+    if address_change.validator_index >= state.validators.len() {
+        return Err(invalid_operation_error(InvalidOperation::ValidatorIndex(
+            address_change.validator_index,
+        )))
+    }
+
+    let withdrawal_credentials_prefix =
+        state.validators[address_change.validator_index].withdrawal_credentials[0];
+
```
ralexstokes, line 57: nitpick
```suggestion
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +40,54 @@ pub fn process_bls_to_execution_change<
...
+    if withdrawal_credentials_prefix != BLS_WITHDRAWAL_PREFIX {
+        return Err(invalid_operation_error(InvalidOperation::WithdrawalCredentialsPrefix(
+            state.validators[address_change.validator_index].withdrawal_credentials[0],
+        )))
+    }
+
+    let public_key: &BlsPublicKey = &address_change.from_bls_public_key;
```
ralexstokes, line 64: drop the type hint?
```suggestion
    let public_key = &address_change.from_bls_public_key;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +40,54 @@ pub fn process_bls_to_execution_change<
...
+
```
ralexstokes, line 65: looks like we dropped the check that the remainder of the withdrawal credentials matches the hash of the `from_bls_pubkey`

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +40,54 @@ pub fn process_bls_to_execution_change<
...
+        Some(state.genesis_validators_root),
+        context,
+    )?;
+    let signed_data =
+        verify_signed_data(&mut address_change.clone(), signature, public_key, domain);
+    if signed_data.is_err() {
+        return Err(invalid_operation_error(InvalidOperation::ExecutionChange(
+            signed_address_change.signature.clone(),
+        )))
+    }
```
ralexstokes, line 78: the API here is that it only returns an error if the signature is invalid, and otherwise just `Ok(())`

so we can do this and even skip bothering w/ the specific error variant
```suggestion
    verify_signed_data(&mut address_change.clone(), signature, public_key, domain)?;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -37,10 +40,54 @@ pub fn process_bls_to_execution_change<
...
+
+    let validator = &mut state.validators[address_change.validator_index];
+
+    validator.withdrawal_credentials[0] = ETH1_ADDRESS_WITHDRAWAL_PREFIX;
+    let mut withdrawal_credentials: Vec<u8> = validator.withdrawal_credentials.as_ref().to_vec();
+    withdrawal_credentials[12..].copy_from_slice(address_change.to_execution_address.as_ref());
+    let withdrawal_credentials: &[u8] = &withdrawal_credentials;
+    let withdrawal_credentials =
+        ByteVector::<32>::try_from(withdrawal_credentials).expect("Wrong size");
+    validator.withdrawal_credentials = withdrawal_credentials;
```
ralexstokes, line 88: see my other comment, you should be able to do what I suggested there now!

I'd just do something like `let withdrawal_credentials = &mut state.validators[address_change.validator_index];` and then do the edits against `withdrawal_credentials`

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -84,6 +84,14 @@ pub enum InvalidOperation {
...
     #[error("invalid execution payload: {0}")]
     ExecutionPayload(#[from] InvalidExecutionPayload),
+    #[error("invalid bls signature for execution chang {0:?}")]
+    ExecutionChange(BlsSignature),
+    #[error("validator index is out of bounds {0}")]
+    ValidatorIndex(usize),
+    #[error("invalid withdrawal credentials prefix{0}")]
+    WithdrawalCredentialsPrefix(u8),
+    #[error("invalid withdrawal credentials public key{0:?}")]
+    WithdrawalCredentialsPublicKey(&'static [u8]),
```
ralexstokes, line 94: its better for managing all of these errors if we add one more level of hierarchy

so let's match the symmetry of the other errors defined here

```suggestion
    #[error("invalid bls signature to execution change: {0}")]
    BlsToExecutionChange(#[from] InvalidBlsToExecutionChange),
```

and then we can put the final variants we need in a new `InvalidBlsToExecutionChange` enum

ralexstokes (changes requested):
nice updates! left some comments

ethereum-consensus/src/capella/block_processing.rs
samtvlabs: Fixed : https://github.com/ralexstokes/ethereum-consensus/pull/251/files#diff-a33e4ef5488fe0396a7616e8ded7baf22a8fec837a613e48edefee587a7c2666R70

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -189,6 +191,18 @@ pub enum InvalidWithdrawals {
     IncorrectWithdrawals { provided: Vec<Withdrawal>, expected: Vec<Withdrawal> },
 }
 
+#[derive(Debug, Error)]
+pub enum InvalidBlsToExecutionChange {
+    #[error("invalid bls signature for execution change {0:?}")]
+    ExecutionChange(BlsSignature),
```
ralexstokes, line 197:
```suggestion
    InvalidSignature(BlsSignature),
```

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -189,6 +191,18 @@ pub enum InvalidWithdrawals {
...
+    #[error("validator index is out of bounds {0}")]
+    BadValidatorIndex(usize),
+    #[error("invalid withdrawal credentials prefix{0}")]
```
ralexstokes, line 200:
```suggestion
    #[error("invalid withdrawal credentials prefix: {0}")]
```

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -189,6 +191,18 @@ pub enum InvalidWithdrawals {
...
+    WithdrawalCredentialsPrefix(u8),
+    #[error("invalid withdrawal credentials public key{0:?}")]
```
ralexstokes, line 202:
```suggestion
    #[error("operation's public key did not match the registered key: {0:?}")]
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,56 @@ pub fn process_bls_to_execution_change<
         BYTES_PER_LOGS_BLOOM,
         MAX_EXTRA_DATA_BYTES,
     >,
-    _signed_address_change: &mut SignedBlsToExecutionChange,
-    _context: &Context,
+    signed_address_change: &SignedBlsToExecutionChange,
+    context: &Context,
 ) -> Result<()> {
-    unimplemented!()
+    let mut address_change = signed_address_change.message.clone();
```
ralexstokes, line 48: no reason to `clone` this here when we can just pass in a `&mut`  like we had

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -189,6 +191,18 @@ pub enum InvalidWithdrawals {
     IncorrectWithdrawals { provided: Vec<Withdrawal>, expected: Vec<Withdrawal> },
 }
 
+#[derive(Debug, Error)]
+pub enum InvalidBlsToExecutionChange {
+    #[error("invalid bls signature for execution change {0:?}")]
+    ExecutionChange(BlsSignature),
+    #[error("validator index is out of bounds {0}")]
+    BadValidatorIndex(usize),
```
ralexstokes, line 199: `Bad` is not that precise....  let's use this instead

also better type safety here to use the `ValidatorIndex` type

```suggestion
    ValidatorIndexOutOfBounds(ValidatorIndex),
```

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -189,6 +191,18 @@ pub enum InvalidWithdrawals {
...
+    #[error("invalid withdrawal credentials prefix{0}")]
+    WithdrawalCredentialsPrefix(u8),
+    #[error("invalid withdrawal credentials public key{0:?}")]
+    WithdrawalCredentialsPublicKey(BlsPublicKey),
```
ralexstokes, line 203: this is a bit better name imo
```suggestion
    PublicKeyMismatch(BlsPublicKey),
```

ralexstokes (changes requested):
nice updates! we are getting closer to merge :)

I left some comments and also I think can tighten up our implementation to drop some use of `clone` and be a bit less verbose

1) keep `signed_address_change` as `&mut SignedBlsToExecutionChange`
you can then immediately do this

```rust
let address_change = &mut signed_address_change.message;
let signature = &signed_address_change.signature;

// ... rest of code
```

2) let's follow [the spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#new-process_bls_to_execution_change) and grab a ref to our validator (it looks like this won't collide anywhere else as far as the borrow checker is concerned)

```rust

let mut validator = &mut state.validators[address_change.validator_index];
```

3) we can then also bind the withdrawal credentials so we don't have to keep going through the `state` data

```rust
let withdrawal_credentials = &validator.withdrawal_credentials;
```

4) let's not `clone` the `from_bls_public_key`,

```rust
let public_key = &address_change.from_bls_public_key;
```

you _will_ need to clone it when assembling the `InvalidBlsToExecutionChange` error, but if you only clone inside the `if` body then we will only pay the cost when necessary

5) I may have misled you w/ the `verify_signed_data` helper as we will run into some issues w/ the borrow checker

instead lets just inline its implementation:

```rust
    let signing_root = compute_signing_root(address_change, domain)?;
    verify_signature(&address_change.from_bls_public_key, signing_root.as_ref(), signature).map_err(Into::into)
```

yash-atreya:
Hey @ralexstokes have made the requested changes in PR #265

samtvlabs:
> #265

@ralexstokes  I have made changes , but want to call out two things 

>let's follow [the spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#new-process_bls_to_execution_change) and grab a ref to our validator (it looks like this won't collide anywhere else as far as the borrow checker is concerned)
let mut validator = &mut state.validators[address_change.validator_index];

IIUC, we already take a reference , and `let mut validator` generates a warning

```shell
warning: variable does not need to be mutable
````

File: ethereum-consensus/src/state_transition/error.rs [hunk 23]
```diff
@@ -189,6 +191,18 @@ pub enum InvalidWithdrawals {
     IncorrectWithdrawals { provided: Vec<Withdrawal>, expected: Vec<Withdrawal> },
 }
 
+#[derive(Debug, Error)]
+pub enum InvalidBlsToExecutionChange {
+    #[error("invalid bls signature for execution change {0:?}")]
+    InvalidSignature(BlsSignature),
+    #[error("validator index is out of bounds {0}")]
```
ralexstokes, line 198:
```suggestion
    #[error("validator index {0} is out of bounds")]
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,57 @@ pub fn process_bls_to_execution_change<
...
+            InvalidBlsToExecutionChange::ValidatorIndexOutOfBounds(address_change.validator_index),
+        )))
+    }
+
+    let validator = &mut state.validators[address_change.validator_index];
+    let withdrawal_credentials = &mut validator.withdrawal_credentials;
+    if withdrawal_credentials[0] != BLS_WITHDRAWAL_PREFIX {
+        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
+            InvalidBlsToExecutionChange::WithdrawalCredentialsPrefix(
+                state.validators[address_change.validator_index].withdrawal_credentials[0],
```
ralexstokes, line 62:
```suggestion
                withdrawal_credentials[0],
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,57 @@ pub fn process_bls_to_execution_change<
...
+            ),
+        )))
+    }
+
+    let public_key = &address_change.from_bls_public_key;
+
```
ralexstokes, line 68:
```suggestion
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,57 @@ pub fn process_bls_to_execution_change<
...
+    if withdrawal_credentials[1..] != hash(public_key.as_slice())[1..] {
```
ralexstokes, line 69:
```suggestion
    if withdrawal_credentials[1..] != hash(public_key.as_ref())[1..] {
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,57 @@ pub fn process_bls_to_execution_change<
...
+
+    let domain = compute_domain(
+        DomainType::BlsToExecutionChange,
+        None,
+        Some(state.genesis_validators_root),
+        context,
+    )?;
+
+    let signing_root = compute_signing_root(address_change, domain)?;
+    verify_signature(&address_change.from_bls_public_key, signing_root.as_ref(), signature)
```
ralexstokes, line 83: this should be fine

```suggestion
    verify_signature(public_key, signing_root.as_ref(), signature)?;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,57 @@ pub fn process_bls_to_execution_change<
...
+        .map_err(|_| {
+            invalid_operation_error(InvalidOperation::BlsToExecutionChange(
+                InvalidBlsToExecutionChange::InvalidSignature(signature.clone()),
+            ))
+        })?;
```
ralexstokes, line 88: this shouldn't be required, let's see how to remove this
```suggestion
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,57 @@ pub fn process_bls_to_execution_change<
...
+
+    validator.withdrawal_credentials[0] = ETH1_ADDRESS_WITHDRAWAL_PREFIX;
+    validator.withdrawal_credentials[1..12].fill(0);
+    validator.withdrawal_credentials[12..]
+        .copy_from_slice(address_change.to_execution_address.as_ref());
```
ralexstokes, line 93: if we are going to take `&mut` on the `withdrawal_credentials` directly, let's just use that

```suggestion
    withdrawal_credentials[0] = ETH1_ADDRESS_WITHDRAWAL_PREFIX;
    withdrawal_credentials[1..12].fill(0);
    withdrawal_credentials[12..]
        .copy_from_slice(address_change.to_execution_address.as_ref());
```

ralexstokes (commented):
nice updates! left some nits below

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,51 @@ pub fn process_bls_to_execution_change<
...
+    }
+
+    let domain = compute_domain(
+        DomainType::BlsToExecutionChange,
+        None,
+        Some(state.genesis_validators_root),
+        context,
+    )?;
+
+    let signing_root = compute_signing_root(address_change, domain)?;
```
ralexstokes, line 81:
```suggestion
        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
            InvalidBlsToExecutionChange::WithdrawalCredentialsPrefix(withdrawal_credentials[0]),
        )))
    }
    
    // NOTE: compute `signing_root` ahead of the public key check to satisfy borrow check
    let domain = compute_domain(
        DomainType::BlsToExecutionChange,
        None,
        Some(state.genesis_validators_root),
        context,
    )?;
    
    let signing_root = compute_signing_root(address_change, domain)?;
    let public_key = &address_change.from_bls_public_key;
    if withdrawal_credentials[1..] != hash(public_key.as_ref())[1..] {
        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
            InvalidBlsToExecutionChange::PublicKeyMismatch(public_key.clone()),
        )))
    }
    
```

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -189,6 +191,18 @@ pub enum InvalidWithdrawals {
     IncorrectWithdrawals { provided: Vec<Withdrawal>, expected: Vec<Withdrawal> },
 }
 
+#[derive(Debug, Error)]
+pub enum InvalidBlsToExecutionChange {
+    #[error("invalid bls signature for execution change {0:?}")]
+    InvalidSignature(BlsSignature),
```
ralexstokes, line 197: we can drop this error variant now

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -1,5 +1,5 @@
 use crate::{
-    capella::Withdrawal,
+    capella::{BlsPublicKey, Withdrawal},
```
ralexstokes, line 2: this should be imported from `primitives`, not `capella`

ralexstokes (changes requested):
let's just clean up the errors and then this should be ready to merge!

ethereum-consensus/src/state_transition/error.rs
```diff
@@ -87,6 +87,8 @@ pub enum InvalidOperation {
     ExecutionPayload(#[from] InvalidExecutionPayload),
     #[error("invalid withdrawals: {0}")]
     Withdrawal(#[from] InvalidWithdrawals),
+    #[error("invalid bls signature to execution change: {0}")]
```
ralexstokes, line 90:
```suggestion
    #[error("invalid BLS signature to execution change: {0}")]
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,49 @@ pub fn process_bls_to_execution_change<
...
+    let signature = &signed_address_change.signature;
+
+    if address_change.validator_index >= state.validators.len() {
+        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
+            InvalidBlsToExecutionChange::ValidatorIndexOutOfBounds(address_change.validator_index),
+        )))
+    }
+
+    let validator = &mut state.validators[address_change.validator_index];
+    let withdrawal_credentials = &mut validator.withdrawal_credentials;
```
ralexstokes, line 58: feels a bit risky to have this &mut Validator lying around

```suggestion
    let withdrawal_credentials = &mut state.validators[address_change.validator_index].withdrawal_credentials;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,49 @@ pub fn process_bls_to_execution_change<
...
+
+    // NOTE: compute `signing_root` ahead of the public key check to satisfy borrow check
+    let domain = compute_domain(
+        DomainType::BlsToExecutionChange,
+        None,
+        Some(state.genesis_validators_root),
+        context,
+    )?;
+
+    let signing_root = compute_signing_root(address_change, domain)?;
```
ralexstokes, line 73:
```suggestion
    let signing_root = compute_signing_root(address_change, domain)?;

```

ralexstokes (approved):
excellent! thanks for staying w/ me on all the back-and-forth

a few minor nits but I think can I merge from here

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,48 @@ pub fn process_bls_to_execution_change<
...
+    let address_change = &mut signed_address_change.message;
+    let signature = &signed_address_change.signature;
+
+    if address_change.validator_index >= state.validators.len() {
+        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
+            InvalidBlsToExecutionChange::ValidatorIndexOutOfBounds(address_change.validator_index),
+        )))
+    }
+
+    let withdrawal_credentials = &mut state.validators[address_change.validator_index].withdrawal_credentials;
```
ralexstokes, line 57:
```suggestion
    let withdrawal_credentials =
         &mut state.validators[address_change.validator_index].withdrawal_credentials;
```

ethereum-consensus/src/capella/block_processing.rs
```diff
@@ -39,10 +42,49 @@ pub fn process_bls_to_execution_change<
...
+    let signature = &signed_address_change.signature;
+
+    if address_change.validator_index >= state.validators.len() {
+        return Err(invalid_operation_error(InvalidOperation::BlsToExecutionChange(
+            InvalidBlsToExecutionChange::ValidatorIndexOutOfBounds(address_change.validator_index),
+        )))
+    }
+
+    let withdrawal_credentials =
+         &mut state.validators[address_change.validator_index].withdrawal_credentials;
```
ralexstokes, line 58:
```suggestion
        &mut state.validators[address_change.validator_index].withdrawal_credentials;
```

PR #261, opened 2023-09-21
disable reqwest default-tls

jxs:
Hi, this allows not pulling the `hyper-tls` and its `openssl-sys` dependency and use `rustls` instead by downstream crates.
Thanks

### DIFF
--- a/beacon-api-client/Cargo.toml
+++ b/beacon-api-client/Cargo.toml
@@ -7,14 +7,15 @@ license = "MIT OR Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["cli"]
+default = ["cli", "reqwest/native-tls"]
 cli = ["clap"]
+rustls = ["reqwest/rustls-tls"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


ralexstokes:
hi @jxs thanks for this!

do you have a link to any description of the problem? I know lighthouse had some issue here so maybe you can just link to a relevant PR?

jxs:
Hi Alex. 
We are currently trying to reduce Lighthouse's compilation time, `hyper-tls` and namely `openssl-sys` (due to static linking of `openssl`) is one of the biggest bottlenecks, one can confirm by running `cargo build --release --timings`. 

After merging https://github.com/sigp/lighthouse/pull/4650 and aligning `reqwest`'s version we can now switch to `rustls` and not depend on 
`hyper-tls`. But `beacon-api-client` depends on `hyper-tls` by default and in a non modifiable way which makes all downstream crates still need to fetch and compile `hyper-tls`.
This PR fixes that by allowing downstream crates to chose their tls flavour and for Lighthouse not pull `hyper-tls` and `openssl-sys` while still offering it as the default.

ralexstokes:
makes sense! thanks for the explanation

PR #266, opened 2023-09-26
allow indirect sub crates to select tls flavor

jxs:
Only noticed after that `mev-rs` also depends on `beacon-api-client` which will also need to expose this option.
Thanks


ralexstokes (approved):
looks fine! do I need to go update `mev-rs` as well?

PR #274, opened 2023-10-01
Remove cfg_attr for serde feature

### LINKED ISSUE #273: remove `cfg_attr` for `serde` feature
I had originally intended to make the `serde` feature optional, and gated the `derive(serde::{Serialize,Deserialize})` attributes behind it, like this https://github.com/ralexstokes/ethereum-consensus/blob/8c204e1a37d7e75d8e067642da4767a4b979282b/ethereum-consensus/src/altair/beacon_block.rs#L11

But it turns out this conditional attribute would also need to apply to each `serde` directive as well, like this https://github.com/ralexstokes/ethereum-consensus/blob/8c204e1a37d7e75d8e067642da4767a4b979282b/ethereum-consensus/src/altair/beacon_block.rs#L44

otherwise, the crate does not compile without the `serde` feature. cf. #191 and #167

in lieu of adding a lot of code to what is already a bit of a noisy codebase (with all the const generics), I think at the moment I favor just ungating the `derive(serde::*)` directives for now and just require `serde`

while it does mean pulling in more crates than could otherwise be possible, `serde` is a common one and is usually in the closure of whatever crates a consumer of this library would already be using, so for most use cases it wouldn't matter. and it doesn't stop us from supporting `no-std` as `serde` has a capability for this as well.

so this issue will be closed once we have made the edits to remove all `cfg_attr(serde)` attributes.

leruaa:
fixes #273 

Is removing `serde` form the `serde` feature enough? Or do you want the `serde` feature removed completely? (the drawback would be to have to set `hex`, `serde_json`, `serde_yaml` non optional...)

Or maybe renaming the feature? It seems a bit wierd to have a `serde` feature without `serde` ;)


ralexstokes (commented):
this is great!

I know it may be annoying but it would be nice to go one step further and consolidate all of the `derive` attributes

so that its just one `derive` attr with the `serde::` stuff at the end of the list

ralexstokes:
> (the drawback would be to have to set hex, serde_json, serde_yaml non optional...)

yeah I like how you did this! let's leave the others as optional and just require the one `serde` crate

I think the name collision is fine but if you want to rename, feel free to make a suggestion :)

leruaa:
@ralexstokes Done :)

ralexstokes (approved):
looks great! thank  you!!

PR #275, opened 2023-10-02
refactor stream_slots with IntervalStream

### LINKED ISSUE #215: refactor `Clock::stream_slots` to use `IntervalStream`
we can likely use something like this to simplify the clock internals:

https://docs.rs/tokio-stream/latest/tokio_stream/wrappers/struct.IntervalStream.html

Perelyn-sama:
to fix #215


ralexstokes (commented):
the idea here was to skip the use of the `async_stream` macro like in #281 

thanks for this PR but ill move ahead with that one for now

PR #281, opened 2023-10-09
implement IntervalStream in Clock::stream_slots

yash-atreya:
Hey @ralexstokes, I picked up #215. 

Refactored using `IntervalStream`

### DIFF
--- a/ethereum-consensus/src/clock.rs
+++ b/ethereum-consensus/src/clock.rs
@@ -195,6 +195,7 @@ impl<T: TimeProvider + Send + Sync> Clock<T> {
 
 pub type SystemClock = Clock<SystemTimeProvider>;
 
+#[cfg(feature = "async")]
 use tokio::time::interval_at;
 #[cfg(feature = "async")]
 use tokio_stream::{wrappers::IntervalStream, Stream, StreamExt};
--- a/ethereum-consensus/src/clock.rs
+++ b/ethereum-consensus/src/clock.rs
@@ -196,19 +196,15 @@ impl<T: TimeProvider + Send + Sync> Clock<T> {
 pub type SystemClock = Clock<SystemTimeProvider>;
 
 #[cfg(feature = "async")]
-use tokio_stream::Stream;
+use tokio_stream::{wrappers::IntervalStream, Stream, StreamExt};
 
 #[cfg(feature = "async")]
 impl<T: TimeProvider + Send + Sync> Clock<T> {
     pub fn stream_slots(&self) -> impl Stream<Item = Slot> + '_ {
-        async_stream::stream! {
-            loop {
-                let slot = self.current_slot().expect("after genesis");
-                yield slot;
-                let duration_until_next_slot = self.duration_until_slot(slot + 1);
-                tokio::time::sleep(duration_until_next_slot).await;
-            }
-        }
+        let slot = self.current_slot().expect("after genesis");
+        let duration_until_next_slot = self.duration_until_slot(slot + 1);
+        let interval_stream = IntervalStream::new(tokio::time::interval(duration_until_next_slot));
+        interval_stream.map(move |_| self.current_slot().expect("after genesis"))
     }
 }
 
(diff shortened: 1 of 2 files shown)


ethereum-consensus/src/clock.rs
```diff
@@ -196,19 +196,15 @@ impl<T: TimeProvider + Send + Sync> Clock<T> {
 pub type SystemClock = Clock<SystemTimeProvider>;
 
 #[cfg(feature = "async")]
-use tokio_stream::Stream;
+use tokio_stream::{wrappers::IntervalStream, Stream, StreamExt};
 
 #[cfg(feature = "async")]
 impl<T: TimeProvider + Send + Sync> Clock<T> {
     pub fn stream_slots(&self) -> impl Stream<Item = Slot> + '_ {
-        async_stream::stream! {
```
ralexstokes, line 204: we can drop the `async-stream` dep now

ethereum-consensus/src/clock.rs
```diff
@@ -196,19 +196,15 @@ impl<T: TimeProvider + Send + Sync> Clock<T> {
...
-            loop {
-                let slot = self.current_slot().expect("after genesis");
-                yield slot;
-                let duration_until_next_slot = self.duration_until_slot(slot + 1);
-                tokio::time::sleep(duration_until_next_slot).await;
-            }
-        }
+        let slot = self.current_slot().expect("after genesis");
+        let duration_until_next_slot = self.duration_until_slot(slot + 1);
+        let interval_stream = IntervalStream::new(tokio::time::interval(duration_until_next_slot));
```
ralexstokes, line 206: this keeps the same amount of offset in the slot for each tick...

I think the best option is to use https://docs.rs/tokio/latest/tokio/time/fn.interval_at.html

So compute the duration until next slot from now, and just have the interval run every `SECONDS_PER_SLOT`, starting from that duration

this changes the API some as we lose a tick for the _current_ slot, but it looks like we would need to implement a custom stream implementation to retain this behavior (would be great if you are up for the challenge :) )

otherwise, lets just make an interval using `interval_at` and add some docs about **not** having a tick for the current slot (which I think is a fine compromise for now...)

ralexstokes (changes requested):
looks good! left some feedback

yash-atreya:
> lets just make an interval using interval_at and add some docs about not having a tick for the current slot (which I think is a fine compromise for now...)

I'll first do this then create a separate PR for the custom stream implementation later?

ralexstokes:
yeah sounds great on moving the custom stream impl to a future PR

ethereum-consensus/src/clock.rs
```diff
@@ -195,20 +195,25 @@ impl<T: TimeProvider + Send + Sync> Clock<T> {
 
 pub type SystemClock = Clock<SystemTimeProvider>;
 
+use tokio::time::interval_at;
```
ralexstokes, line 198: we should also gate this
```suggestion
#[cfg(feature = "async")]
use tokio::time::interval_at;
```

ethereum-consensus/src/clock.rs
```diff
@@ -195,20 +195,25 @@ impl<T: TimeProvider + Send + Sync> Clock<T> {
...
-                tokio::time::sleep(duration_until_next_slot).await;
-            }
-        }
+        let slot = self.current_slot().expect("after genesis");
+        let duration_until_next_slot = self.duration_until_slot(slot + 1);
+
+        // Stream will not emit the `current_slot`, it starts from the next slot. This is until a custom stream implementation is added to fix this. Ref: https://github.com/ralexstokes/ethereum-consensus/pull/281#discussion_r1350932968
+        let interval = interval_at(
+            tokio::time::Instant::now() + duration_until_next_slot,
+            Duration::from_secs(configs::mainnet::SECONDS_PER_SLOT),
```
ralexstokes, line 213: this should be parametric over the `Clock`s `SECONDS_PER_SLOT`

```suggestion
            Duration::from_secs(self.seconds_per_slot),
```

ralexstokes (changes requested):
nice updates! a few things to fix and then this should be ready to merge

ralexstokes:
closing in lieu of #385

Issue #283, opened 2023-10-10
add "no-default-features" build to CI to ensure things work w/o default feature set


ralexstokes:
we should have a build step like this in the CI to ensure that things also build w/o default features

https://github.com/ralexstokes/ssz-rs/blob/main/.github/workflows/ci.yml#L37

and it looks like `beacon-api-client` currently does not succeed!

yash-atreya:
I can take this

yash-atreya:
> @yash-atreya are you still working on this ? happy to take it over the line

Please go ahead.

ralexstokes:
I think we just need to do this: https://github.com/ralexstokes/ethereum-consensus/pull/284/files#r1355734863

to get that PR ready to merge -- @samtvlabs if you want to make a pass at it, feel free, otherwise I can just do in the next few days

samtvlabs:
Apologies dropped this ball on this will get this over the line this weekend

PR #284, opened 2023-10-11
ci: build using --no-default-features

yash-atreya:
Addresses #283


.github/workflows/ci.yml
```diff
@@ -33,6 +33,9 @@ jobs:
       - name: Build
         run: cargo build --all-targets --all-features --verbose
 
+      - name: Build `no-std`
```
ralexstokes, line 36: would be nice to mirror this in the `justfile` as well :)

https://github.com/ralexstokes/ssz-rs/blob/main/justfile#L9

PR #314, opened 2023-11-06
Create verify_blob_sidecar_inclusion_proof()

EchoAlice:
WIP

### DIFF
--- a/ethereum-consensus/src/deneb/blob_sidecar.rs
+++ b/ethereum-consensus/src/deneb/blob_sidecar.rs
@@ -51,10 +51,9 @@ fn generalized_index_for_blob_index(i: usize) -> Result<GeneralizedIndex, Merkle
     BeaconBlockBody::generalized_index(path)
 }
 
-// TODO: Overflow bugs? Refactor to be more concise?
-fn get_subtree_index(i: usize) -> usize {
-    let floorlog2 = (i as f64).log2().floor();
-    ((i as i32) % 2_i32.pow(floorlog2 as u32)) as usize
+fn get_subtree_index(i: GeneralizedIndex) -> usize {
+    let floorlog2 = i.checked_ilog2().expect("gindex isn't zero");
+    i % 2usize.pow(floorlog2)
 }
 
 #[derive(
--- a/ethereum-consensus/src/deneb/blob_sidecar.rs
+++ b/ethereum-consensus/src/deneb/blob_sidecar.rs
@@ -73,7 +73,7 @@ pub fn verify_blob_sidecar_inclusion_proof<
     let depth = KZG_COMMITMENT_INCLUSION_PROOF_DEPTH;
     let root = signed_block_header.message.body_root;
 
-    Ok(is_valid_merkle_branch(leaf, branch, depth, subtree_index as usize, root)?)
+    Ok(is_valid_merkle_branch(leaf, branch, depth, subtree_index, root)?)
 }
 
 fn generalized_index_for_blob_index<
@@ -113,11 +113,8 @@ fn generalized_index_for_blob_index<
     >::generalized_index(path)
 }
 
-fn get_subtree_index(i: GeneralizedIndex) -> Result<u32, MerkleizationError> {
-    match i.checked_ilog2() {
-        Some(floorlog2) => Ok(floorlog2),
-        None => Err(MerkleizationError::InvalidGeneralizedIndex),
-    }
+fn get_subtree_index(i: GeneralizedIndex) -> Result<usize, MerkleizationError> {
+    i.checked_ilog2().map(|index| index as usize).ok_or(MerkleizationError::InvalidGeneralizedIndex)
 }
 
 #[derive(
--- a/ethereum-consensus/src/deneb/blob_sidecar.rs
+++ b/ethereum-consensus/src/deneb/blob_sidecar.rs
@@ -1,5 +1,6 @@
 use crate::{
     deneb::{
+        mainnet::BeaconBlockBody,
         polynomial_commitments::{KzgCommitment, KzgProof},
         SignedBeaconBlockHeader,
     },
@@ -28,23 +29,32 @@ pub struct BlobSidecar<
     pub kzg_commitment_inclusion_proof: Vector<Bytes32, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>,
 }
 
+// TODO: Convert method to free-standing function
 impl<const BYTES_PER_BLOB: usize, const KZG_COMMITMENT_INCLUSION_PROOF_DEPTH: usize>
     BlobSidecar<BYTES_PER_BLOB, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>
 {
     pub fn verify_blob_sidecar_inclusion_proof(&mut self) -> Result<(), Error> {
-        // TODO: Calculate real gindex
-        let index = 4;
+        let blob_index = generalized_index_for_blob_index(self.index)?;
+        let subtree_index = get_subtree_index(blob_index);
+
         let leaf = self.kzg_commitment.hash_tree_root()?;
         let branch = self.kzg_commitment_inclusion_proof.as_ref();
         let depth = KZG_COMMITMENT_INCLUSION_PROOF_DEPTH;
         let root = self.signed_block_header.message.body_root;
 
-        Ok(is_valid_merkle_branch(leaf, branch, depth, index, root)?)
+        Ok(is_valid_merkle_branch(leaf, branch, depth, subtree_index, root)?)
     }
 }
 
-fn generalized_index_for_blob_index(index: usize) -> usize {
-    57
+fn generalized_index_for_blob_index(i: usize) -> Result<GeneralizedIndex, MerkleizationError> {
+    let path = &["blob_kzg_commitments".into(), i.into()];
+    BeaconBlockBody::generalized_index(path)
+}
+
+// TODO: Overflow bugs? Refactor to be more concise?
+fn get_subtree_index(i: usize) -> usize {
+    let floorlog2 = (i as f64).log2().floor();
+    ((i as i32) % 2_i32.pow(floorlog2 as u32)) as usize
 }
 
 #[derive(
--- a/ethereum-consensus/src/deneb/blob_sidecar.rs
+++ b/ethereum-consensus/src/deneb/blob_sidecar.rs
@@ -43,6 +43,10 @@ impl<const BYTES_PER_BLOB: usize, const KZG_COMMITMENT_INCLUSION_PROOF_DEPTH: us
     }
 }
 
+fn generalized_index_for_blob_index(index: usize) -> usize {
+    57
+}
+
 #[derive(
     Default, Debug, Clone, SimpleSerialize, PartialEq, Eq, serde::Serialize, serde::Deserialize,
 )]
--- a/ethereum-consensus/src/deneb/blob_sidecar.rs
+++ b/ethereum-consensus/src/deneb/blob_sidecar.rs
@@ -45,11 +45,9 @@ pub fn verify_blob_sidecar_inclusion_proof<
     const MAX_BLS_TO_EXECUTION_CHANGES: usize,
     const MAX_BLOB_COMMITMENTS_PER_BLOCK: usize,
     const KZG_COMMITMENT_INCLUSION_PROOF_DEPTH: usize,
+    const BYTES_PER_BLOB: usize,
 >(
-    index: BlobIndex,
-    kzg_commitment: &mut KzgCommitment,
-    signed_block_header: &SignedBeaconBlockHeader,
-    kzg_commitment_inclusion_proof: &Vector<Bytes32, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>,
+    blob_sidecar: &mut BlobSidecar<BYTES_PER_BLOB, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>,
 ) -> Result<(), Error> {
     let g_index = generalized_index_for_blob_index::<
         MAX_PROPOSER_SLASHINGS,
@@ -66,15 +64,15 @@ pub fn verify_blob_sidecar_inclusion_proof<
         MAX_WITHDRAWALS_PER_PAYLOAD,
         MAX_BLS_TO_EXECUTION_CHANGES,
         MAX_BLOB_COMMITMENTS_PER_BLOCK,
-    >(index)?;
+    >(blob_sidecar.index)?;
     let subtree_index = get_subtree_index(g_index)?;
 
-    let leaf = kzg_commitment.hash_tree_root()?;
-    let branch = kzg_commitment_inclusion_proof.as_ref();
+    let leaf = blob_sidecar.kzg_commitment.hash_tree_root()?;
+    let branch = blob_sidecar.kzg_commitment_inclusion_proof.as_ref();
     let depth = KZG_COMMITMENT_INCLUSION_PROOF_DEPTH;
-    let root = signed_block_header.message.body_root;
+    let root = blob_sidecar.signed_block_header.message.body_root;
 
-    Ok(is_valid_merkle_branch(leaf, branch, depth, subtree_index, root)?)
+    is_valid_merkle_branch(leaf, branch, depth, subtree_index, root).map_err(Into::into)
 }
 
 fn generalized_index_for_blob_index<
@@ -93,7 +91,7 @@ fn generalized_index_for_blob_index<
     const MAX_BLS_TO_EXECUTION_CHANGES: usize,
     const MAX_BLOB_COMMITMENTS_PER_BLOCK: usize,
 >(
-    i: usize,
+    i: BlobIndex,
 ) -> Result<GeneralizedIndex, MerkleizationError> {
     let path = &["blob_kzg_commitments".into(), i.into()];
     BeaconBlockBody::<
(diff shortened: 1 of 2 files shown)


ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +27,28 @@ pub struct BlobSidecar<
...
+    pub fn verify_blob_sidecar_inclusion_proof(&mut self) -> bool {
+        // TODO: Calculate real gindex
+        let gindex = 4;
+
+        // TODO: Handle hash_tree_root() error
+        let leaf = &self.kzg_commitment.hash_tree_root().unwrap();
+        let branch: Vec<_> = self
+            .kzg_commitment_inclusion_proof
+            .iter()
+            .map(|bytes| Node::try_from(bytes.as_slice()).unwrap())
```
ralexstokes, line 42:
```suggestion
            .map(|bytes| Node::try_from(bytes.as_slice()).expect("correct size input"))
```

ralexstokes, line 42: nitpick
```suggestion
            .map(|bytes| Node::try_from(bytes.as_ref()).unwrap())
```

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +27,28 @@ pub struct BlobSidecar<
...
+            .collect();
+        let depth = KZG_COMMITMENT_INCLUSION_PROOF_DEPTH;
```
ralexstokes, line 44: move this to be a `pub const` in this file plz

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +27,28 @@ pub struct BlobSidecar<
...
+        let index = gindex;
```
ralexstokes, line 45:
```suggestion
```

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +27,28 @@ pub struct BlobSidecar<
     pub kzg_commitment_inclusion_proof: Vector<Bytes32, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>,
 }
 
+impl<const BYTES_PER_BLOB: usize, const KZG_COMMITMENT_INCLUSION_PROOF_DEPTH: usize>
+    BlobSidecar<BYTES_PER_BLOB, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>
+{
+    pub fn verify_blob_sidecar_inclusion_proof(&mut self) -> bool {
+        // TODO: Calculate real gindex
+        let gindex = 4;
```
ralexstokes, line 35: lets just define index once, this may change a bit after I land the ssz-rs updates
```suggestion
        let index = 4;
```

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +27,28 @@ pub struct BlobSidecar<
     pub kzg_commitment_inclusion_proof: Vector<Bytes32, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>,
 }
 
+impl<const BYTES_PER_BLOB: usize, const KZG_COMMITMENT_INCLUSION_PROOF_DEPTH: usize>
+    BlobSidecar<BYTES_PER_BLOB, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>
+{
+    pub fn verify_blob_sidecar_inclusion_proof(&mut self) -> bool {
```
ralexstokes, line 33: let's use the crate-wide error for this, and make a new error type for this specific set of failure modes

imports up above:
```rust
use crate::Error;
```

```suggestion
    pub fn verify_blob_sidecar_inclusion_proof(&mut self) -> Error {
```

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +27,28 @@ pub struct BlobSidecar<
...
+        // TODO: Calculate real gindex
+        let gindex = 4;
+
+        // TODO: Handle hash_tree_root() error
+        let leaf = &self.kzg_commitment.hash_tree_root().unwrap();
```
ralexstokes, line 38: w/ error changes then:
```suggestion
        let leaf = &self.kzg_commitment.hash_tree_root()?;
```

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +27,28 @@ pub struct BlobSidecar<
...
+        let branch: Vec<_> = self
+            .kzg_commitment_inclusion_proof
+            .iter()
+            .map(|bytes| Node::try_from(bytes.as_slice()).unwrap())
+            .collect();
+        let depth = KZG_COMMITMENT_INCLUSION_PROOF_DEPTH;
+        let index = gindex;
+        let root = &self.signed_block_header.message.body_root;
+
+        is_valid_merkle_branch(leaf, branch.iter(), depth, index, root)
```
ralexstokes, line 48: I actually think I'll update this API some and include a bespoke error type, rather than just `bool`

we can make the changes once I land this and the g-index updates to `ssz-rs`

ralexstokes (commented):
great start! left some comments, and this otherwise will be blocked on some `ssz-rs` updates ill try to get to soon(tm)

ralexstokes:
@EchoAlice FYI updated the `is_valid_merkle_branch` API, so if you rebase from `main`, you can simplify your use of it here; check out that PR to see other places it changed for an example

ethereum-consensus/src/deneb/blob_sidecar.rs
EchoAlice: That's why they pay you the big bucks

ralexstokes (commented):
nice updates! we need to implement the logic to compute the g_index (or could compute in pyspec and just copy over as a constant

EchoAlice:
> nice updates! we need to implement the logic to compute the g_index (or could compute in pyspec and just copy over as a constant

Which option makes more sense to you?

ralexstokes:
unblocked on the generalized index computation with #324

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -1,5 +1,6 @@
 use crate::{
     deneb::{
+        mainnet::BeaconBlockBody,
```
ralexstokes, line 3: you want to use the generic type here, not a particular preset

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -28,23 +29,32 @@ pub struct BlobSidecar<
...
 
-fn generalized_index_for_blob_index(index: usize) -> usize {
-    57
+fn generalized_index_for_blob_index(i: usize) -> Result<GeneralizedIndex, MerkleizationError> {
+    let path = &["blob_kzg_commitments".into(), i.into()];
+    BeaconBlockBody::generalized_index(path)
+}
+
+// TODO: Overflow bugs? Refactor to be more concise?
+fn get_subtree_index(i: usize) -> usize {
```
ralexstokes, line 55: just following the spec to the letter....
```suggestion
fn get_subtree_index(i: GeneralizedIndex) -> usize {
```

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -28,23 +29,32 @@ pub struct BlobSidecar<
...
+    let floorlog2 = (i as f64).log2().floor();
```
ralexstokes, line 56: the spec avoids floating point arithmetic for a number of reasons, one of them being determinism of execution

I'd just use this instead https://doc.rust-lang.org/std/primitive.usize.html#method.checked_ilog2

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -28,23 +29,32 @@ pub struct BlobSidecar<
...
+    ((i as i32) % 2_i32.pow(floorlog2 as u32)) as usize
```
ralexstokes, line 57: just a note that `i as i32` would truncate the number, tossing away the upper half

ralexstokes, line 57: you should just be able to pipe the floor log 2 value into this

https://doc.rust-lang.org/std/primitive.usize.html#method.pow

ralexstokes (changes requested):
nice updates! left a few more comments

ethereum-consensus/src/deneb/blob_sidecar.rs
EchoAlice: Yep.  I *assumed* we wouldn't be using giant numbers.  But understand that's probably bad form

ralexstokes: well its more that you don't know

your input is a 64-bit uint, so you should program assuming that

ralexstokes: types are useful *because* they communicate contracts

ethereum-consensus/src/deneb/blob_sidecar.rs
EchoAlice: i'm not sure how to import a generic BeaconBlockBody

ralexstokes:
```suggestion
        beacon_block::BeaconBlockBody,
```

you'll need to provide the const generic arguments which is why I was suggesting just having this be a free-standing function, rather than having types all over the place

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +29,33 @@ pub struct BlobSidecar<
...
+    }
+}
+
+fn generalized_index_for_blob_index(i: usize) -> Result<GeneralizedIndex, MerkleizationError> {
+    let path = &["blob_kzg_commitments".into(), i.into()];
+    BeaconBlockBody::generalized_index(path)
+}
+
+fn get_subtree_index(i: GeneralizedIndex) -> usize {
+    let floorlog2 = i.checked_ilog2().expect("gindex isn't zero");
```
ralexstokes, line 55: we should prefer to return an error, rather than panicking, which `expect` would do here if `i == 0`

instead, we can do `unwrap_or` and return this error: https://github.com/ralexstokes/ssz-rs/blob/6329986d14e71a98a35cc9323f88cbef0332b6dd/ssz-rs/src/merkleization/mod.rs#L38

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -29,21 +29,21 @@ pub struct BlobSidecar<
...
-// TODO: Convert method to free-standing function
-impl<const BYTES_PER_BLOB: usize, const KZG_COMMITMENT_INCLUSION_PROOF_DEPTH: usize>
-    BlobSidecar<BYTES_PER_BLOB, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>
-{
-    pub fn verify_blob_sidecar_inclusion_proof(&mut self) -> Result<(), Error> {
-        let blob_index = generalized_index_for_blob_index(self.index)?;
-        let subtree_index = get_subtree_index(blob_index);
+pub fn verify_blob_sidecar_inclusion_proof<const KZG_COMMITMENT_INCLUSION_PROOF_DEPTH: usize>(
+    index: BlobIndex,
+    mut kzg_commitment: KzgCommitment,
```
ralexstokes, line 34: better to take `&mut` here so we don't run around cloning this data all the time

ethereum-consensus/src/deneb/blob_sidecar.rs
EchoAlice: I'm not sure how to provide the const generic arguments.  When i add the const generics to the appropriate function signatures, i get this error:

```    Checking ethereum-consensus v0.1.1 (/home/echo/Desktop/Ethereum/ethereum-consensus/ethereum-consensus)
error[E0282]: type annotations needed
  --> ethereum-consensus/src/deneb/blob_sidecar.rs:81:5
   |
81 |     BeaconBlockBody::generalized_index(path)
   |     ^^^^^^^^^^^^^^^ cannot infer the value of const parameter `MAX_PROPOSER_SLASHINGS` declared on the struct `BeaconBlockBody`
```

ethereum-consensus/src/deneb/blob_sidecar.rs
EchoAlice: `.unwrap_or()` expects a `u32` as input instead of an error.  I'm going explicitly handle the returned value of `.checked_ilog2()` with a `match` statement.  Thoughts?

ralexstokes: sorry I meant this: https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -26,6 +28,98 @@ pub struct BlobSidecar<
...
+        MAX_EXTRA_DATA_BYTES,
+        MAX_BYTES_PER_TRANSACTION,
+        MAX_TRANSACTIONS_PER_PAYLOAD,
+        MAX_WITHDRAWALS_PER_PAYLOAD,
+        MAX_BLS_TO_EXECUTION_CHANGES,
+        MAX_BLOB_COMMITMENTS_PER_BLOCK,
+    >::generalized_index(path)
+}
+
+// TODO: Return an error if checked_ilog2 returns "None"
```
ralexstokes, line 116:
```suggestion
```

EchoAlice: Oops

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -26,6 +28,94 @@ pub struct BlobSidecar<
     pub signed_block_header: SignedBeaconBlockHeader,
     pub kzg_commitment_inclusion_proof: Vector<Bytes32, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>,
 }
+pub fn verify_blob_sidecar_inclusion_proof<
```
ralexstokes, line 31:
```suggestion
}

pub fn verify_blob_sidecar_inclusion_proof<
```

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +29,95 @@ pub struct BlobSidecar<
...
+    const MAX_TRANSACTIONS_PER_PAYLOAD: usize,
+    const MAX_WITHDRAWALS_PER_PAYLOAD: usize,
+    const MAX_BLS_TO_EXECUTION_CHANGES: usize,
+    const MAX_BLOB_COMMITMENTS_PER_BLOCK: usize,
+    const KZG_COMMITMENT_INCLUSION_PROOF_DEPTH: usize,
+>(
+    index: BlobIndex,
+    kzg_commitment: &mut KzgCommitment,
+    signed_block_header: &SignedBeaconBlockHeader,
+    kzg_commitment_inclusion_proof: &Vector<Bytes32, KZG_COMMITMENT_INCLUSION_PROOF_DEPTH>,
```
ralexstokes, line 52: ok, one last thing I think!

let's mirror [the spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#verify_blob_sidecar_inclusion_proof) and take a `BlobSidecar` as an input!

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +29,95 @@ pub struct BlobSidecar<
...
+    const SYNC_COMMITTEE_SIZE: usize,
+    const BYTES_PER_LOGS_BLOOM: usize,
+    const MAX_EXTRA_DATA_BYTES: usize,
+    const MAX_BYTES_PER_TRANSACTION: usize,
+    const MAX_TRANSACTIONS_PER_PAYLOAD: usize,
+    const MAX_WITHDRAWALS_PER_PAYLOAD: usize,
+    const MAX_BLS_TO_EXECUTION_CHANGES: usize,
+    const MAX_BLOB_COMMITMENTS_PER_BLOCK: usize,
+>(
+    i: usize,
```
ralexstokes, line 96: either inline this function, or let's keep the typing
```suggestion
    i: BlobIndex,
```

ethereum-consensus/src/deneb/blob_sidecar.rs
```diff
@@ -27,6 +29,95 @@ pub struct BlobSidecar<
...
+        MAX_BLOB_COMMITMENTS_PER_BLOCK,
+    >(index)?;
+    let subtree_index = get_subtree_index(g_index)?;
+
+    let leaf = kzg_commitment.hash_tree_root()?;
+    let branch = kzg_commitment_inclusion_proof.as_ref();
+    let depth = KZG_COMMITMENT_INCLUSION_PROOF_DEPTH;
+    let root = signed_block_header.message.body_root;
+
+    Ok(is_valid_merkle_branch(leaf, branch, depth, subtree_index, root)?)
```
ralexstokes, line 77: nit but I think this is easier to read, rather than the dense Ok() and ? usage

```suggestion
    is_valid_merkle_branch(leaf, branch, depth, subtree_index, root).map_error(Into::into)
```

ralexstokes (changes requested):
one more round then we can merge :)

ethereum-consensus/src/deneb/blob_sidecar.rs
EchoAlice: i assume you meant `.map_err(Into::into)`, right?

EchoAlice:
> perfect! thanks for all the back and forth :)

Thanks for all of the guidance!

PR #318, opened 2023-11-10
fix for HOLESKY_GENESIS_TIME

Owen66:
The current holesky time stamp is wrong.

For example the timestamp of slot [308297](https://holesky.beaconcha.in/slot/308297) is 1699601964.
Using the correct genesis time we have 1695902400 + (308297 * 12) == 1699601964.
However, using the timestamp in the repo atm we have 1694786400 + (308297 * 12) == 1698485964.


ralexstokes (approved):
thank you! this never got updated after the holesky reboot

PR #319, opened 2023-11-22
feat: enable default traits for validator types 

Doozers:
helpful derives while working with the crates :)


Doozers:
@ralexstokes

ralexstokes (approved):
makes sense, thanks for the PR!

PR #320, opened 2023-12-07
fix: `rustls` feature for beacon api client

rkrasiuk:
## Description

Fix `rustls` feature by toggling `tls` features on `mev-share-sse` dependency.

Before:
```sh
➜ cargo tree -p beacon-api-client --no-default-features --features rustls --target all | grep openssl
│   │   │   │   ├── openssl v0.10.61
│   │   │   │   │   ├── openssl-macros v0.1.1 (proc-macro)
│   │   │   │   │   └── openssl-sys v0.9.97
│   │   │   │   ├── openssl-probe v0.1.5
│   │   │   │   ├── openssl-sys v0.9.97 (*)
```

After:
```sh
➜ cargo tree -p beacon-api-client --no-default-features --features rustls --target all | grep openssl
<no output>
```

### DIFF
--- a/beacon-api-client/Cargo.toml
+++ b/beacon-api-client/Cargo.toml
@@ -9,8 +9,8 @@ license = "MIT OR Apache-2.0"
 [features]
 default = ["cli", "native-tls"]
 cli = ["clap"]
-rustls = ["reqwest/rustls-tls"]
-native-tls = ["reqwest/default-tls"]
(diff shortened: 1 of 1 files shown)


ralexstokes (approved):
ah nice catch, thank you!

I wonder if there is a way to test for this, other than something along the lines of `cargo tree` like you posted...

PR #323, opened 2024-02-01
Update Light Client Structs for Capella and Deneb

ec2:


ralexstokes:
thanks for this! it seems like there is some crossed wires w/ the cargo version and the formatter though...

if I run `cargo +nightly fmt --all` locally, it doesn't add all of these changes to the formatting

using `rustfmt 1.7.0-nightly (b11fbfbf 2024-02-03)`

maybe your local setup wasn't detecting the local `rustfmt.toml`?

also, I can only find one file to add support for `capella` definitions -- did you also mean to include `deneb` changes?

I'd consider opening a new PR with just the changes for `capella`, and not running the linter without the prescribed config

ralexstokes:
thanks for this! however, I've addressed this in #346

PR #378, opened 2024-04-09
feat: add auxiliary deposit functions

### LINKED ISSUE #375: Deposit refactoring in `phase0`
There are some additional helpers added in the `consensus-specs` for deposit handling (like `apply_deposit`).

Would be nice to sync this repo w/ those updates

https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#deposits

Some missing functions we could add:

`add_validator_to_registry`
`apply_deposit`

jacobkaufmann:
resolves #375


jacobkaufmann:
can remove generated `electra` spec if desired

ethereum-consensus/src/phase0/block_processing.rs
```diff
@@ -311,6 +317,100 @@ pub fn get_validator_from_deposit(deposit: &Deposit, context: &Context) -> Valid
...
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    public_key: &BlsPublicKey,
+    withdrawal_credentials: &Bytes32,
+    amount: u64,
```
ralexstokes, line 376: imo spec should say `Gwei` here, and we can do this here either way for a dash of type-safety

ethereum-consensus/src/phase0/block_processing.rs
```diff
@@ -311,6 +317,100 @@ pub fn get_validator_from_deposit(deposit: &Deposit, context: &Context) -> Valid
...
+        ETH1_DATA_VOTES_BOUND,
+        VALIDATOR_REGISTRY_LIMIT,
+        EPOCHS_PER_HISTORICAL_VECTOR,
+        EPOCHS_PER_SLASHINGS_VECTOR,
+        MAX_VALIDATORS_PER_COMMITTEE,
+        PENDING_ATTESTATIONS_BOUND,
+    >,
+    public_key: BlsPublicKey,
+    withdrawal_credentials: Bytes32,
+    amount: u64,
```
ralexstokes, line 342: update to `Gwei` if you change below

ethereum-consensus/src/altair/block_processing.rs
```diff
@@ -159,6 +156,103 @@ pub fn process_attestation<
...
+        public_key.clone(),
+        withdrawal_credentials.clone(),
+        amount,
+        context,
+    );
+
+    Ok(())
+}
+
 pub fn process_deposit<
```
ralexstokes, line 256: I think we should also be able to delete this `process_deposit` definition and let the `spec-gen` handle it

ethereum-consensus/src/phase0/block_processing.rs
```diff
@@ -311,6 +317,100 @@ pub fn get_validator_from_deposit(deposit: &Deposit, context: &Context) -> Valid
...
+    let signing_root = compute_signing_root(&mut deposit_message, domain)?;
+    if verify_signature(public_key, signing_root.as_ref(), signature).is_err() {
+        // NOTE: explicitly return with no error and also no further mutations to `state`
+        return Ok(());
+    }
+
+    add_validator_to_registry(
+        state,
+        public_key.clone(),
+        withdrawal_credentials.clone(),
```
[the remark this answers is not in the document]
jacobkaufmann: yea, looks like we could take from `deposit_message`, but ultimately `get_validator_from_deposit` needs ownership, which is why I went with owned value in `add_validator_to_registry`.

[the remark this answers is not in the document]
ralexstokes: yeah, which is why we should be able to take them away from `deposit_message`

generally I pass by ref until the "leaf" of the AST but passing by value is fine here

ethereum-consensus/src/phase0/block_processing.rs
```diff
@@ -290,16 +292,20 @@ pub fn process_attestation<
     Ok(())
 }
 
-pub fn get_validator_from_deposit(deposit: &Deposit, context: &Context) -> Validator {
-    let amount = deposit.data.amount;
+pub fn get_validator_from_deposit(
+    public_key: BlsPublicKey,
+    withdrawal_credentials: Bytes32,
+    amount: u64,
```
ralexstokes, line 298: also `Gwei`?

ethereum-consensus/src/phase0/block_processing.rs
```diff
@@ -309,7 +315,96 @@ pub fn get_validator_from_deposit(deposit: &Deposit, context: &Context) -> Valid
     }
 }
 
-pub(crate) const DEPOSIT_MERKLE_DEPTH: usize = DEPOSIT_CONTRACT_TREE_DEPTH + 1;
+pub const DEPOSIT_MERKLE_DEPTH: usize = DEPOSIT_CONTRACT_TREE_DEPTH + 1;
```
ralexstokes, line 318: is there a reason visibility changed here?

ralexstokes (approved):
looks great! two questions

ethereum-consensus/src/phase0/block_processing.rs
jacobkaufmann: it was not being included in the subsequent `spec` modules without that change.

ralexstokes: lets just get rid of this, https://github.com/ralexstokes/ethereum-consensus/pull/379

ralexstokes (approved):
looks great!

PR #405, opened 2024-10-08
Add conversion between ByteVector and FixedBytes

Akagi201:
It will help to work with alloy crates.

MSRV >= 1.81

The latest alloy-primitives crate requires rustc >= 1.79, and reth requires rustc >= 1.81

https://github.com/alloy-rs/core/blob/main/Cargo.toml#L8
https://github.com/paradigmxyz/reth/blob/main/Cargo.toml#L4

### DIFF
--- a/ethereum-consensus/src/crypto/bls.rs
+++ b/ethereum-consensus/src/crypto/bls.rs
@@ -3,6 +3,7 @@ use crate::serde::try_bytes_from_hex_str;
 use crate::{primitives::Bytes32, ssz::prelude::*};
 use blst::{min_pk as bls_impl, BLST_ERROR};
 use sha2::{Digest, Sha256};
+use tracing::error;
 use std::{
     fmt,
     ops::{Deref, DerefMut},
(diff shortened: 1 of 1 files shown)


ralexstokes:
do you have a use case in mind, e.g. another repo or project that needs the conversion?

I have been thinking for some time to base the primitives here around `alloy` to the extent possible, although we run into some blocks with deriving the relevant SSZ implementations

also bumping the MSRV is an important change so it would be nice to either surface that in this PR (or even make a separate PR)

Akagi201:
In `alloy::rpc::types::beacon::BlsPublicKey;`

```rust
/// BLS signature type
pub type BlsSignature = FixedBytes<BLS_SIGNATURE_BYTES_LEN>;
```

In `ethereum_consensus::primitives::BlsPublicKey;

```rust
pub struct PublicKey(ByteVector<BLS_PUBLIC_KEY_BYTES_LEN>);
```

I face the issue that I need to convert between the above two types in my project. Also It will be better that we can only use alloy primitives types, maybe it's not easy.

The latest alloy-primitives crate requires rustc >= 1.79, and reth requires rustc >= 1.81

https://github.com/alloy-rs/core/blob/main/Cargo.toml#L8
https://github.com/paradigmxyz/reth/blob/main/Cargo.toml#L4

Akagi201:
Also I found that the ci lint errors are not related to my PR changes.

Issue #415, opened 2025-03-28
Conflict upgrading alloy to 0.13.0 due to ethereum-consensus pinning older c-kzg


estensen:
Alloy upgrades c-kzg to v2 in [v0.13.0](https://github.com/alloy-rs/alloy/releases/tag/v0.13.0)

However, this library currently pins c-kzg to v1. Because we depend on both ethereum-consensus (which requires v1) and alloy (which now requires v2), we’re blocked from upgrading to alloy 0.13.0

PR #416, opened 2025-04-03
bump c_kzg version and reth deps

SozinM:
We need to bump version, to keep this compatible with the latest reth

Thanks for @mattsse for tips 

fixes https://github.com/ralexstokes/ethereum-consensus/issues/415


ralexstokes:
@SozinM thank you for this!

i tried to fix the lint error here: https://github.com/ralexstokes/ethereum-consensus/commit/d5aeaeb60ee9dd1ad3a57948aefc23e1bd0e036f

but could not push to the right place; i think you'll need to do on your end

ethereum-consensus/src/deneb/presets/mod.rs
```diff
@@ -7,4 +7,9 @@ pub struct Preset {
     pub max_blobs_per_block: usize,
 }
 
-pub const TRUSTED_SETUP_JSON: &str = include_str!("trusted_setup.json");
+/// Precompute value that optimizes computing cell kzg proofs.
+///
+/// Set to 0 as we do not use `compute_cells_and_kzg_proofs` or `recover_cells_and_kzg_proofs`.
+///
+/// Learn more: <https://github.com/ethereum/c-kzg-4844/blob/dffa18ee350aeef38f749ffad24a27c1645fb4f8/README.md?plain=1#L112>
+pub const PRECOMPUTE: u64 = 0;
```
ralexstokes, line 15: i think this should live in `ethereum-consensus/src/crypto/kzg.rs` for now

ethereum-consensus/src/deneb/presets/mod.rs
```diff
@@ -7,4 +7,9 @@ pub struct Preset {
     pub max_blobs_per_block: usize,
 }
 
-pub const TRUSTED_SETUP_JSON: &str = include_str!("trusted_setup.json");
```
ralexstokes, line 10: it looks like we no longer need to load from JSON here as the relevant data is provided from `c-kzg` now

can you remove this `trusted_setup.json` file as well?

ethereum-consensus/src/bin/ec/blobs/bundler.rs
```diff
@@ -41,7 +41,7 @@ pub fn bundle(blobs: Vec<Blob>, kzg_settings: &spec::KzgSettings) -> Result<Blob
 
 // Assumes a serde_json-encoded array of `Vec<Blob>` on `reader` and uses the mainnet trusted setup.
 pub fn from_reader(reader: impl Read) -> Result<BlobsBundle, Error> {
-    let kzg_settings = spec::kzg_settings_from_json(TRUSTED_SETUP_JSON)?;
+    let kzg_settings = c_kzg::ethereum_kzg_settings(PRECOMPUTE);
```
ralexstokes, line 44: i'd like to encapsulate `c_kzg` inside the `crypto` module in this library

to that end, let's keep a helper function like `kzg_settings_from_json` which instead can be something like `kzg_settings_with_precompute(precompute: u64)`

this is nice as we will want to use a different `precompute` value for fulu and later forks, so each of those versions of the spec can customize via this `kzg_settings_with_precompute` function

to handle the clone issue, either have just this function, and wrap in an Arc (like it was in `ethereum-consensus/src/state_transition/context.rs`), or make a `kzg_setings_with_precompute_arc` function

i think i have a slight preference for just the one constructor

ralexstokes (changes requested):
this is great! thanks for updating this

i have a few comments but then should be ready to merge

PR #420, opened 2025-07-18
fix: rejecting mathematically valid BLS12-381 private keys

barnabasbusa:
The Bug: The ethereum-consensus library's SecretKey::try_from() method was overly strict, rejecting mathematically valid BLS12-381 private keys that exceed the field order instead of performing proper modular reduction.

### DIFF
--- a/ethereum-consensus/src/crypto/bls.rs
+++ b/ethereum-consensus/src/crypto/bls.rs
@@ -192,8 +192,27 @@ impl TryFrom<&[u8]> for SecretKey {
     type Error = Error;
 
     fn try_from(data: &[u8]) -> Result<Self, Self::Error> {
-        let inner = bls_impl::SecretKey::from_bytes(data).map_err(BLSTError::from)?;
-        Ok(Self(inner))
+        // First try direct construction (for keys already in valid range)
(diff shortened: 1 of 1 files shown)


ethereum-consensus/src/crypto/bls.rs
```diff
@@ -192,8 +192,27 @@ impl TryFrom<&[u8]> for SecretKey {
...
+        // If direct construction fails, perform modular reduction per EIP-2333
+        // This handles keys that are valid but exceed the field order
+        if data.len() == BLS_SECRET_KEY_BYTES_LEN {
+            // Convert bytes to big integer
+            let mut key_int = [0u8; 32];
+            key_int.copy_from_slice(data);
+            
+            // Use HKDF with the key as input material to perform proper reduction
+            // This follows EIP-2333 key generation pattern
+            let sk = bls_impl::SecretKey::key_gen(&key_int, &[]).map_err(BLSTError::from)?;
```
ralexstokes, line 209: if we want to go this route, we probably want something like this

https://docs.rs/blst/0.3.15/blst/min_pk/struct.SecretKey.html#method.derive_master_eip2333

let me see if there is a better way to expose this though before making any changes

ralexstokes:
closing in lieu of #421

